Issue #18, opened 2022-04-24
[BUG] tensorrt package missing leading to onnx-to-trt failure 


pdmack:
```
# morpheus tools onnx-to-trt \
>             --input_model /common/models/sid-minibert-onnx/1/model.onnx \
>             --output_model /common/models/sid-minibert-trt_b1-8_b1-16_b1-32.engine \
>             --batches 1 8 \
>             --batches 1 16 \
>             --batches 1 32 \
>             --seq_length 256 \
>             --max_workspace_size 4000
Traceback (most recent call last):
  File "/opt/conda/envs/morpheus/bin/morpheus", line 11, in <module>
    sys.exit(run_cli())
  File "/opt/conda/envs/morpheus/lib/python3.8/site-packages/morpheus/cli.py", line 1381, in run_cli
    cli(obj={}, auto_envvar_prefix='CLX', show_default=True, prog_name="morpheus")
  File "/opt/conda/envs/morpheus/lib/python3.8/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/opt/conda/envs/morpheus/lib/python3.8/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/opt/conda/envs/morpheus/lib/python3.8/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/opt/conda/envs/morpheus/lib/python3.8/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/opt/conda/envs/morpheus/lib/python3.8/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/conda/envs/morpheus/lib/python3.8/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/opt/conda/envs/morpheus/lib/python3.8/site-packages/morpheus/cli.py", line 136, in new_func
    return f(ctx, *args, **kwargs)
  File "/opt/conda/envs/morpheus/lib/python3.8/site-packages/morpheus/cli.py", line 233, in onnx_to_trt
    gen_engine(c)
  File "/opt/conda/envs/morpheus/lib/python3.8/site-packages/morpheus/utils/onnx_to_trt.py", line 30, in gen_engine
    import tensorrt as trt
[...]

PR #23, opened 2022-04-25
TensorRT installation warning for onnx_to_trt

cwharris:
Contributes to #18

### DIFF
--- a/morpheus/cli.py
+++ b/morpheus/cli.py
@@ -217,6 +217,12 @@ def tools(ctx: click.Context, **kwargs):
 @prepare_command()
 def onnx_to_trt(ctx: click.Context, **kwargs):
 
+    try:
+        import tensorrt
+    except:
+        logger.error("onnx_to_trt requires the tensorrt module and runtime to be installed.")
+        return
+
     logger.info("Generating onnx file")
 
     # Convert batches to a list
--- a/morpheus/cli.py
+++ b/morpheus/cli.py
@@ -217,12 +217,6 @@ def tools(ctx: click.Context, **kwargs):
 @prepare_command()
 def onnx_to_trt(ctx: click.Context, **kwargs):
 
-    try:
-        import tensorrt
-    except:
-        logger.error("onnx_to_trt requires the tensorrt module and runtime to be installed.")
-        return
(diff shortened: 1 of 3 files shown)


GPUtester:
Can one of the admins verify this patch?

pdmack:
LGTM

pdmack (commented):
LGTM

morpheus/cli.py
```diff
@@ -217,6 +217,12 @@ def tools(ctx: click.Context, **kwargs):
 @prepare_command()
 def onnx_to_trt(ctx: click.Context, **kwargs):
 
+    try:
+        import tensorrt
+    except:
+        logger.error("onnx-to-trt requires the tensorrt module and runtime to be installed.")
+        return
+
```
mdemoret-nv, line 225: Can we actually move this import check into the `onnx_to_trt.py` file? This check should really be done there and in case the utility is used outside of the CLI. Whis would allow us to remove the existing `import tensorrt as trt` that is in the `gen_engine` function and replace it with this code at the top level.

mdemoret-nv (changes requested):

morpheus/utils/onnx_to_trt.py
```diff
@@ -27,7 +32,12 @@ def gen_engine(c: ConfigOnnxToTRT):
     """
 
     # Local imports to avoid requiring TensorRT to generate the docs
-    import tensorrt as trt
+    try:
+        import tensorrt as trt
+    except:
+        logger.error("onnx_to_trt requires the tensorrt module and runtime to be installed.")
+        return
+
```
mdemoret-nv, line 40: Can you move this to the top of the file? It doesnt need to be inside the function. You should add
```python
autodoc_mock_imports = [
    "morpheus.cli",  # Dont document the CLI in Sphinx
    "tqdm",
    "tensorrt",  # Ignore loading tensorrt
]
```
to `docs/source/conf.py` to prevent this from being loaded as part of the sphinx autodoc

mdemoret-nv (changes requested):

mdemoret-nv (approved):

PR #28, opened 2022-04-25
Log parsing example updates

efajardo-nv:
~- Add missing vocabulary file (`bert-base-cased-vocab.txt`)~
- Fix path to run script in example README
- Use `os.cpu_count()` instead of `psutil.cpu_count()` for pipeline config's `num_threads`

### DIFF
--- a/data/bert-base-cased-vocab.txt
+++ b/data/bert-base-cased-vocab.txt
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:94f958e06f58dbca894d1fff97e59242d2dbf188ce776a08dee9c7faf8096211
+size 219226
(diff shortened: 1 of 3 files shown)


GPUtester:
Can one of the admins verify this patch?

cwharris (approved):

mdemoret-nv (changes requested):
If we are adding `data/bert-base-cased-vocab.txt` we should also add the matching `data/bert-base-uncased-vocab.txt`.

cwharris:
@mdemoret-nv `data/bert-base-uncased-vocab.txt` isn't used in this example. Should these data files be consolidated somewhere? It looks like we have them copy pasted in a few locations. The files aren't identical, though, so it's confusing.

mdemoret-nv (approved):

PR #29, opened 2022-04-25
Removing `no_args_is_help` from CLI commands

### LINKED ISSUE #26: [BUG] Remove `no_args_is_help` from CLI
**Describe the bug**
If you use a morpheus CLI command without any arguments or options, it just prints the help instead of adding that stage

For example, this launch command will just print the help:

```bash
(morpheus)$ morpheus run pipeline-nlp \
                from-file --filename=data/pcap_dump.jsonlines \
                deserialize
Configuring Pipeline via CLI
Usage: morpheus run pipeline-nlp deserialize [OPTIONS]

Options:
  --help  Show this message and exit.
```

**Expected behavior**
The pipeline should run with the specified commands

mdemoret-nv:
Closes #26

### DIFF
--- a/morpheus/cli.py
+++ b/morpheus/cli.py
@@ -306,6 +306,7 @@ def install(**kwargs):
 def run(ctx: click.Context, **kwargs):
 
     # Since the option isnt the same name as `should_use_cpp` anymore, manually set the value here.
+    CppConfig.set_should_use_cpp(kwargs.pop("use_cpp", CppConfig.get_should_use_cpp()))
 
     pass
 
--- a/morpheus/cli.py
+++ b/morpheus/cli.py
@@ -52,7 +52,6 @@ def str_to_file_type(file_type_str: str):
 
 command_kwargs = {
     "context_settings": dict(show_default=True, ),
-    "no_args_is_help": True,
 }
 
 ALIASES = {
@@ -125,8 +124,6 @@ def prepare_command(parse_config: bool = False):
 
             kwargs = _without_empty_args(kwargs)
 
-            CppConfig.set_should_use_cpp(kwargs.pop("use_cpp", CppConfig.get_should_use_cpp()))
-
             # Apply the config if desired
             if parse_config:
                 config = get_config_from_ctx(ctx)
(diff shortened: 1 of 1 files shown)


morpheus/cli.py
```diff
@@ -125,8 +124,6 @@ def new_func(*args, **kwargs):
 
             kwargs = _without_empty_args(kwargs)
 
-            CppConfig.set_should_use_cpp(kwargs.pop("use_cpp", CppConfig.get_should_use_cpp()))
```
cwharris, line 128: This seems unrelated. Is this being removed intentionally, and why?

morpheus/cli.py
```diff
@@ -306,7 +310,11 @@ def run(ctx: click.Context, **kwargs):
     pass
 
 
-@click.group(chain=True, short_help="Run the inference pipeline with a NLP model", cls=AliasedGroup, **command_kwargs)
+@click.group(chain=True,
+             short_help="Run the inference pipeline with a NLP model",
+             no_args_is_help=True,
```
cwharris, line 315: does "args" include the stages, such as `from-file`, or just the `--` flags? If it's just the flags, then this might be problematic, since in the issue we use
```
morpheus run pipeline-nlp \
             from-file --filename=data/pcap_dump.jsonlines \
             deserialize
```
where pipeline-nlp has no flags. But I assume `from-file` counts as an "arg", and this is fine.

cwharris (changes requested):

dagardner-nv (approved):
LGTM

morpheus/cli.py
mdemoret-nv: This only applies to the last command in the chain. So this would not trigger the help for `pipeline-nlp`:
```
morpheus run pipeline-nlp \
             from-file --filename=data/pcap_dump.jsonlines \
             deserialize
```
But this would:
```
morpheus run pipeline-nlp
```
Since `morpheus run pipeline-nlp` would be an invalid pipeline with no commands, it would be better to just print the help info. The `no_args_is_help` option should be enabled for all `@cli.group` decorators, especially ones with `@cli.group(chain=True)`

morpheus/cli.py
mdemoret-nv: This was a merge error which I tried to fix this this PR at the same time but I forgot to add it back into the `run` command (Good catch because I had the fix wrong).

This value should only be set in the `run` function which actually has the `--use_cpp` option. Setting it in `prepare_command` will just set the same value over and over again.

cwharris (approved):

PR #50, opened 2022-04-26
Adding option to specify a conda channel alias during conda-build

mdemoret-nv:
If there are 404 errors for the channels `nvidia` or `rapidsai`, then the conda-build scripts can fail. This adds the option to specify `CONDA_CHANNEL_ALIAS` before calling `./ci/conda/recipes/run_conda_build.sh`.

Since we are currently getting these errors, the option has been enabled by default in the docker container builds but will not be used unless manually specified.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -114,7 +114,11 @@ COPY docker/conda/environments/cuda${CUDA_VER}_dev.yml ./docker/conda/environmen
 # Update the morpheus environment
 RUN --mount=type=bind,from=conda_bld_deps,source=/opt/conda/conda-bld,target=/opt/conda/conda-bld \
     --mount=type=cache,id=conda_pkgs,target=/opt/conda/pkgs,sharing=locked \
+    # Temp add channel_alias to get around conda 404 errors
+    conda config --env --set channel_alias http://conda-mirror.gpuci.io/ &&\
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -128,7 +129,8 @@ RUN --mount=type=ssh \
     --mount=type=cache,id=workspace_cache,target=/workspace/.cache,sharing=locked \
     --mount=type=cache,id=conda_pkgs,target=/opt/conda/pkgs,sharing=locked \
     source activate base &&\
-    MORPHEUS_ROOT=/workspace CONDA_BLD_DIR=/opt/conda/conda-bld CONDA_ARGS="--no-test" ./ci/conda/recipes/run_conda_build.sh morpheus
+    # Temp add CONDA_CHANNEL_ALIAS to get around conda-build 404 errors
```
pdmack, line 132: Tech debt?

pdmack (commented):
LGTM
but maybe need @dagardner-nv @drobison00 @cwharris also?

docker/Dockerfile
mdemoret-nv: Ya, its not great but hopefully this wont be a permanent thing. Worst case scenario is we use these mirrors for 22.04.

drobison00:
Looks fine for now, so long as we circle back on the conda package build commands.

drobison00 (approved):

BartleyR (approved):

PR #53, opened 2022-04-26
Fixes bug in WriteToFileStage when configured with a filename without a path

### LINKED ISSUE #48: [BUG] WriteToFileStage fails when given a filename without a path
**Describe the bug**
Initializing WriteToFileStage with:
`WriteToFileStage(config, filename='test.json')`

Will cause a failure in `_build_single`:
`os.makedirs(os.path.dirname(self._output_file), exist_ok=True)`
The call to `os.path.dirname` will return '' causing `os.mkdirs` to raise an exception

dagardner-nv:
fixes #48

### DIFF
--- a/morpheus/pipeline/output/to_file.py
+++ b/morpheus/pipeline/output/to_file.py
@@ -120,7 +120,7 @@ class WriteToFileStage(SinglePortStage):
             def node_fn(input: neo.Observable, output: neo.Subscriber):
 
                 # Ensure our directory exists
-                os.makedirs(os.path.dirname(self._output_file), exist_ok=True)
+                os.makedirs(os.path.realpath(os.path.dirname(self._output_file)), exist_ok=True)
 
                 # Open up the file handle
                 with open(self._output_file, "a") as out_file:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -172,9 +172,7 @@ def chdir_tmpdir(request: pytest.FixtureRequest, tmp_path):
     """
     Executes a test in the tmp_path directory
     """
-    print(tmp_path)
     os.chdir(tmp_path)
-    print(1)
     yield
     os.chdir(request.config.invocation_dir)
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -167,6 +167,18 @@ def reload_modules(request: pytest.FixtureRequest):
             importlib.reload(mod)
 
 
+@pytest.fixture(scope="function")
+def chdir_tmpdir(request: pytest.FixtureRequest, tmp_path):
+    """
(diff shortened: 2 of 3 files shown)


morpheus/pipeline/output/to_file.py
```diff
@@ -120,7 +120,7 @@ def _build_single(self, seg: neo.Segment, input_stream: StreamPair) -> StreamPai
             def node_fn(input: neo.Observable, output: neo.Subscriber):
 
                 # Ensure our directory exists
-                os.makedirs(os.path.dirname(self._output_file), exist_ok=True)
+                os.makedirs(os.path.realpath(os.path.dirname(self._output_file)), exist_ok=True)
```
mdemoret-nv, line 123: Isnt this backwards. Shouldnt it be 
```
os.makedirs(os.path.dirname(os.path.realpath(self._output_file)), exist_ok=True)
```

tests/conftest.py
```diff
@@ -167,6 +167,18 @@ def reload_modules(request: pytest.FixtureRequest):
...
 
 
+@pytest.fixture(scope="function")
+def chdir_tmpdir(request: pytest.FixtureRequest, tmp_path):
+    """
+    Executes a test in the tmp_path directory
+    """
+    print(tmp_path)
+    os.chdir(tmp_path)
+    print(1)
```
mdemoret-nv, line 177: Dont need these prints

mdemoret-nv (changes requested):

morpheus/pipeline/output/to_file.py
mdemoret-nv: This leads me to believe that your test might not be working as intended

dagardner-nv: As far as I can tell the output is always the same, if you prefer one over the other let me know.

```
In [1]: import os

In [2]: os.path.realpath(os.path.dirname('README.md'))
Out[2]: '/home/dagardner/work/morpheus'

In [3]: os.path.realpath(os.path.dirname('README.md')) == os.path.dirname(os.path.realpath('README.md'))
Out[3]: True

In [4]: os.path.realpath(os.path.dirname('./README.md')) == os.path.dirname(os.path.realpath('./README.md'))
Out[4]: True

In [5]: os.path.realpath(os.path.dirname('./examples/log_parsing/README.md')) == os.path.dirname(os.path.realpath('./examples/log_parsin
   ...: g/README.md'))
Out[5]: True
```

tests/conftest.py
dagardner-nv: Well now :) that is embarrassing now isn't it :) 
good catch

mdemoret-nv (approved):

PR #58, opened 2022-04-27
Fix type-o in log parsing example & small formatting fix

dagardner-nv:


dagardner-nv:
I removed unused SubwordTokenizer per feedback from Eli

examples/log_parsing/preprocessing.py
```diff
@@ -80,7 +79,6 @@ def __init__(self,
         self._do_lower_case = do_lower_case
         self._add_special_tokens = add_special_tokens
 
-        self._tokenizer: SubwordTokenizer = None
```
BartleyR, line 83: How are we doing the tokenization now, or did we just move this to a different location?

BartleyR: Disregard, I saw your comment above.

PR #61, opened 2022-04-27
Fix the nlp_si_detection example

### LINKED ISSUE #59: [DOC] nlp_si_detection example needs to be updated
## Report incorrect documentation
* Old triton
* Reference to git sub module 
* Remove vizualization flag
* In cpp mode it fails an assert early `Check failed: this->mess_count == this->count At this time, mess_count and count must be the same for slicing`
* in python mode it fails later with `ValueError: Cannot align indices with non-unique values`

**Location of incorrect documentation**
`examples/nlp_si_detection`

dagardner-nv:
I was getting a failed assert with the original pipeline, I added `--truncation=True` to the pre-process stage which caused the assert to go away, can someone confirm this is correct?
```
Check failed: this->mess_count == this->count At this time, mess_count and count must be the same for slicing
```

* Update version fo triton
* Remove ref to git submodule
* Fix path to labels file
* Remove buffer stages
* Remove vizualization flag
* Add `--truncation=True` to the preprocess stage
* Morpheus stops on its own now

fixes #59

### DIFF
--- a/examples/nlp_si_detection/README.md
+++ b/examples/nlp_si_detection/README.md
@@ -70,21 +70,17 @@ Below is a visualization of the pipeline showing all of the stages and data type
 
 ## Setup
 
-This example utilizes the Triton Inference Server to perform inference. The neural network model is provided in a separate submodule repo. Make sure the Morpheus Models repo is checked out with:
-
-```bash
-git submodule update --init --recursive
-```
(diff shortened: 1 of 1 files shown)


mdemoret-nv (approved):

mdemoret-nv:
@gpucibot merge

PR #62, opened 2022-04-27
Ensure default path values in morpheus.cli are no longer relative

### LINKED ISSUE #60: [BUG] Relative paths in the CLI should always be computed as relative to the root
**Describe the bug**
Some of the default values for command line flags are relative paths ex:
```
@click.option('--labels_file', default="data/labels_nlp.txt",
```

The problem then is that this only works when run from the root of the repo.  Several of our own examples are run from within the example sub-directory.

dagardner-nv:
Moves the `data` dir under the morpheus dir. This allows it to be easily installable via setuptools & discoverable relative to `morpheus.__file__` and sets a new attribute `morpheus.DATA_DIR`.

This change also fixes a few random bugs related to non-inplace morpheus builds.

Allows morphues cli to be run are outside of the root of the repo
fixes #60


morpheus/cli.py
```diff
@@ -41,7 +42,7 @@
 # List all of the options in from morpheus._lib.file_types.FileTypes without importing the object. This slows down
 # autocomplete too much.
 FILE_TYPE_NAMES = ["auto", "csv", "json"]
-
+MORPHEUS_ROOT = os.environ.get('MORPHEUS_ROOT', os.path.dirname(os.path.dirname(morpheus.__file__)))
```
cwharris, line 45: Cool. We could eventually do something in our examples. :+1:

tests/test_cli.py
```diff
@@ -189,6 +190,11 @@ def test_pipeline_ae(self, config, callback_values):
         config.ae.userid_column_name = "user_col"
         config.ae.userid_filter = "user321"
 
+        with open(os.path.join(TEST_DIRS.data_dir, 'columns_ae.txt')) as fh:
+            expected_columns = [line.strip() for line in fh]
```
cwharris, line 194: We use this three times. should we move it in to a helper function?
```python
def read_lines_from_test_data(filename):
    with open(os.path.join(TEST_DIRS.data_dir, filename)) as fh:
            expected_columns = [line.strip() for line in fh]
```
```python
expected_columns = read_lines_from_test_data("columns_ae.txt")
```

cwharris (changes requested):

tests/test_cli.py
dagardner-nv: done

tests/test_cli.py
```diff
@@ -189,6 +197,9 @@ def test_pipeline_ae(self, config, callback_values):
         config.ae.userid_column_name = "user_col"
         config.ae.userid_filter = "user321"
 
+        expected_columns = self._read_data_file(os.path.join(TEST_DIRS.data_dir, 'columns_ae.txt'))
```
cwharris, line 200: will we (most) always be loading from TEST_DIRS? If so, maybe we should move the path join in to `_read_data_file`.

cwharris (changes requested):

examples/nlp_si_detection/README.md
```diff
@@ -108,13 +108,13 @@ morpheus --debug --log_level=DEBUG \
    `# Run a pipeline with 8 threads and a model batch size of 32 (Must match Triton config)` \
    run --num_threads=8 --pipeline_batch_size=1024 --model_max_batch_size=32 \
    `# Specify a NLP pipeline with 256 sequence length (Must match Triton config)` \
-   pipeline-nlp --model_seq_length=256 --labels_file=$MORPHEUS_ROOT/data/labels_nlp.txt \
```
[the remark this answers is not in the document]
dagardner-nv: It was setting the default value. Previously, due to this bug, the default values only worked if you were in the root of the repo. 

Explicitly setting this flag is no longer requited.

tests/test_cli.py
dagardner-nv: I would really rather not, there are other dirs that unittests interact with like the `tests/expected_data` dir and the `tmp_path` so I wouldn't want to hinder the method, potentially making it less flexible in the future.

examples/nlp_si_detection/README.md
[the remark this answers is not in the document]
cwharris: That's what I suspected, but I'm wondering... should we update all of the default values to use real files/values?

Seems out of scope for this PR. Just curious what your thoughts are.

cwharris (approved):
LGTM

examples/nlp_si_detection/README.md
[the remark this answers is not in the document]
dagardner-nv: I updated the ones in the examples dir. However most of the examples have their own run.py script and don't use the morpheus.cli

morpheus/cmake/utils/python_module_tools.cmake
```diff
@@ -196,26 +196,30 @@ macro(add_python_module MODULE_NAME)
   # succeed
   add_dependencies(all_python_targets ${TARGET_NAME})
 
-  # Before installing, create the custom command to generate the stubs
-  set(pybind11_stub_file "${MODULE_NAME}/__init__.pyi")
+  if (MORPHEUS_PYTHON_INPLACE_BUILD)
```
mdemoret-nv, line 199: We still want to be building the stub files for non-inplace builds. Non-inplace builds should still generate stub files since they need to be packaged up with the python wheel.

mdemoret-nv (changes requested):
Couple of small changes

CMakeLists.txt
```diff
@@ -136,5 +137,10 @@ install(
     PATTERN "py.typed"
 )
 
+install(
+  DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/morpheus/data
+  DESTINATION ${MORPHEUS_PY_INSTALL_DIR}/morpheus
+  COMPONENT Wheel
+)
```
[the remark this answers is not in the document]
dagardner-nv: done

mdemoret-nv (approved):
Looks good now.

mdemoret-nv:
@gpucibot merge

PR #68, opened 2022-04-27
Update docker build to support adding a non-root default user 

### LINKED ISSUE #22: [FEA] Add ability to build containers that default to a user other than root.
See title

**Describe alternatives you've considered**

**Additional context**

drobison00:
Resolves #22 

Now running

```
DOCKER_EXTRA_ARGS="--build-arg MORPHEUS_USER=[YOUR_USER_NAME]" ./docker/build_container_dev.sh
```
Will create a new default user for the development container, with sudo privileges.


cwharris:
What's the purpose of this? I would have expected a non-root "morpheus" user, but I'm confused why it is useful to specify an arbitrary username at container build time via an arg.

drobison00:
@cwharris
Running as root within a container will stop you from mounting X11 into the container or easy GUI applications, I use it for debugging.

Another significant issue is that it is really annoying to run a dev container and have it create files in your current host user directory as the 'root' user. This allows someone to create the container with a user name/id that matches the user id of the user launching the container, so the permissions align.

docker/Dockerfile
```diff
@@ -149,6 +149,23 @@ RUN --mount=type=ssh \
...
+      && apt-get update \
+      && apt-get install -y sudo \
+      && apt-get clean \
+      && echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers ; \
+    else \
+      echo "Skipping user creation..." ;\
+    fi
+
+# Allow default user to execute sudo commands in container
+RUN echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
```
mdemoret-nv, line 167: Shouldnt this go in the `else` branch of the above if statement? Instead of the line `echo "Skipping user creation..." ;\`. This would save a layer

docker/Dockerfile
```diff
@@ -172,13 +189,29 @@ COPY "./models" "./models"
...
+    useradd -ms /workspace ${MORPHEUS_USER} \
+      && usermod -aG sudo ${MORPHEUS_USER} \
+      && apt-get update \
+      && apt-get install -y sudo \
+      && apt-get clean \
+      && echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers ; \
+    else \
+      echo "Skipping user creation..." ;\
+    fi
+
```
mdemoret-nv, line 214: Can we move this into one of the above layers (such as base or conda_env) to avoid duplicating it in both the `development` and `runtime` targets? We dont need to set the `USER` until needed but we certainly can create the user in the base image.

CONTRIBUTING.md
```diff
@@ -90,7 +90,13 @@ This workflow utilizes a docker container to set up most dependencies ensuring a
...
       ```
       Would build the container `morpheus:my_tag`.
-   1. Note: This does not build any Morpheus or Neo code and defers building the code until the entire repo can be mounted into a running container. This allows for faster incremental builds during development.
+   1. If you would like to build the container to have a default user, other than root, add the following to your 
+   `build_container_dev.sh` call.
+      ```bash
+      DOCKER_EXTRA_ARGS="--build-arg MORPHEUS_USER=[YOUR_USER_NAME]" ./docker/build_container_dev.sh
+      ```
+   1. Note: This does not build any Morpheus or Neo code and defers building the code until the entire repo can be 
+   1. mounted into a running container. This allows for faster incremental builds during development.
```
mdemoret-nv, line 99:
```suggestion
   1. Note: This does not build any Morpheus or Neo code and defers building the code until the entire repo can be mounted into a running container. This allows for faster incremental builds during development.
```

CONTRIBUTING.md
```diff
@@ -90,7 +90,13 @@ This workflow utilizes a docker container to set up most dependencies ensuring a
       DOCKER_IMAGE_TAG=my_tag ./docker/build_container_dev.sh
       ```
       Would build the container `morpheus:my_tag`.
-   1. Note: This does not build any Morpheus or Neo code and defers building the code until the entire repo can be mounted into a running container. This allows for faster incremental builds during development.
+   1. If you would like to build the container to have a default user, other than root, add the following to your 
+   `build_container_dev.sh` call.
+      ```bash
+      DOCKER_EXTRA_ARGS="--build-arg MORPHEUS_USER=[YOUR_USER_NAME]" ./docker/build_container_dev.sh
+      ```
```
mdemoret-nv, line 97: I think this is pretty safe to include by default in the `build_container_dev.sh` and omit from `build_container_release.sh` is it not? We could just add the following to `build_container_dev.sh`:

```
DOCKER_EXTRA_ARGS="${DOCKER_EXTRA_ARGS} --build-arg MORPHEUS_USER=$USER"
```

Or better yet, make a new option `DOCKER_IMAGE_USER` in `build_container.sh` which will set the new build arg.

mdemoret-nv (changes requested):
This looks like a nice improvement. Have a few suggestions. Before merging, I would like to get @pdmack to weigh in on this one regarding any packaging/security concerns this may raise.

mdemoret-nv:
Also, it's worth noting that the Docker best practices guide cautions against installing sudo in a container and suggests an alternative: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user

docker/Dockerfile
drobison00: Yes, this makes sense. I'll move it up.

docker/Dockerfile
drobison00: Yeah, good catch. I'll move it into the check.

CONTRIBUTING.md
drobison00: Generally speaking, it probably makes sense to specify a user for the release container too, just from a security standpoint, since we do things like mount volumes by default. We could set it to something default, like 'morpheus'. I don't think there is any downside to doing so.

mdemoret-nv:
Closing since this is out of date. Can reopen with updated code.

pdmack:
Gonna poke at this again

pdmack:
Might create another PR based on 22.11

PR #100, opened 2022-05-10
Add pybind11/cython stubs to Morpheus package

### LINKED ISSUE #99: [FEA] Add Python Stubs for Cython/PyBind11 Modules
Python modules created with cython/pybind11 do not show up in IDEs and fail all linting which is a poor developer experience.

Utilize mypy or pybind11-stubgen to create pyi stubs for all C++ generated code. This allows linters and IDEs to pickup and check python code using these modules

mdemoret-nv:
This MR allows python classes created in C++ to show up in IDEs and work with linters. It works by using pybind11-stubgen to create .pyi files that are then included in the package. The result is this:

image

Only downside is the pybind11-stubgen isnt too actively maintained (but its also not complex). And the option for generating the output location is limited. So in our output we get:

```
common/
   __init__.pyi
cudf_helpers/
   __init__.pyi
messages/
   __init__.pyi
stages/
   __init__.pyi
common.so
cudf_helpers.so
messages.so
stages.so
```
It would be better to be organized like the following:
```
common.pyi
common.so
cudf_helpers.pyi
cudf_helpers.so
messages.pyi
messages.so
stages.pyi
stages.so
```

The only alternative is to use mypy. However, they dont support docstrings according to this issue: https://github.com/python/mypy/issues/11965

Closes #99


drobison00 (approved):
LGTM -- If its not terribly complex, is it worth us forking the library and adding the functionality we want?

Also, is morpheus/py.typed supposed to be empty?

mdemoret-nv:
> LGTM -- If its not terribly complex, is it worth us forking the library and adding the functionality we want?

We could. That wouldnt be terribly complex. Might be better to just upstream a new option to do what we want. I'm not terribly bothered by the layout, but it could be cleaner.

> Also, is morpheus/py.typed supposed to be empty?

Yes, that is the specification.

BartleyR (approved):

mdemoret-nv:
Changed the label to breaking since this could break existing scripts, depending on how the import was done.

mdemoret-nv:
@gpucibot merge

PR #98, opened 2022-05-10
Reorganize the python package files

mdemoret-nv:
This refactors all of the python package files to improve the directory structure, separate classes into their own files, improve readability and reduce merge conflicts. The main morpheus package will now be subdivided into the following submodules:

```
morpheus/
   _lib/
   io/
   messages/
   pipelines/
   stages/
      general/
      inference/
      input/
      output/
      postprocess/
      preprocess
   utils/
```

The largest differences are coming from separating the `morpheus/pipeline.py` into separate files and breaking all stages into their own package (resulting in quite a few new files).

Remaining questions:
- [x] How to handle package reloads for `morpheus.messages`
- [ ] What to call utilities? `utils`? or `utilities`?

Closes #98

PR #105, opened 2022-05-13
Update example imports after python package reorg

efajardo-nv:
Few more imports in following examples needed updating after python package reorg (#98).
- log parsing
- abp pcap detection


pdmack:
@efajardo-nv kind of looks like (to my naked eye) @dagardner-nv fixed these in #62 ?

PR #116, opened 2022-05-19
Add script to capture triton config

pdmack:
Utility script to render a "starter" Triton config from onnx, tf or trt "bare" models.

Mix of bash, jq, sed, and awk. Over time could be replaced by python with proper json and protobuf libraries. Best effort attempt to preserve gpu, cpu accelerator and optional parameters fields if the model in fact has its own config.pbtxt. 

From the help:
```
This script retrieves a computed model configuration from NVIDIA Triton. The Triton
Inference Server can be launched with "--strict-model-config=false" which means it
will create a minimal Triton configuration (config.pbtxt) from the required
elements of a model if one has not been provided.

This only applies to TensorRT, TensorFlow saved-model, and ONNX models. Once the
initial configuration file is derived, it is expected that optional and advanced
settings will be applied by hand.

By default, the script will take the JSON output from Triton and reformat it
to stdout for usage as a new configuration (i.e., save stdout to config.pbtxt).

Developed and tested with the 22.04 release of Triton using Morpheus ONNX and FIL
sample models only.

For more information on Triton model configuration, please refer to the online docs:
https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md
https://github.com/triton-inference-server/common/blob/main/protobuf/model_config.proto

This script is intended for initial bootstrapping of a basic Triton model configuration.
Consult the Triton Model Navigator for advanced model optimization tooling.
https://github.com/triton-inference-server/model_navigator/blob/main/docs/optimize_for_triton.md
```

### DIFF
--- a/scripts/get-triton-config.sh
+++ b/scripts/get-triton-config.sh
@@ -1,0 +1,241 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: Copyright (c) 2021-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


mdemoret-nv (approved):

mdemoret-nv:
@gpucibot merge

PR #139, opened 2022-06-02
Limit which lfs assets are pulled by default

### LINKED ISSUE #88: [FEA] Improve LFS with Scripts and Optional Download
Currently, Git LFS is used to store large and binary files in the repo. However, some users have not had success with Git LFS. By default, Git LFS will download all files for a particular commit, even if the user is unlikely to need every file. This adds about 2 GB to the repo checkout.

Ideally, the LFS configuration would be improved to minimize the download size for the average user. Utility scripts could be added to download the necessary files for a particular workload/example/model. This would improve the user experience by minimizing the repo checkout size and would not require anyone to learn the necessary Git LFS commands to download ignored or skipped files.

**Describe alternatives you've considered**
Other options include storing large files external to Git (i.e. S3) and building a set of scripts or tools to download these files as necessary. This option (while preferred by some) would just trade one problem for another. We would need to develop our own versioning system for storing these large files, create the tools to download them, and maintain a system which functions very similarly to Git LFS but is slightly different. While Git LFS is far from perfect, we can mitigate some of the issues by correctly configuring our repo and providing scripts to shield users from needing to deal with Git LFS.

**Additional context**
The recommended solution from Rapids Ops is outlined here: https://docs.rapids.ai/resources/git/

dagardner-nv:
* Adds helper script `scripts/fetch_data.py` to pull additional assets.
* Update `README.md` & `CONTRIBUTING.md` docs to reflect changes
* Updated the hammah unittest to no longer depend on training data.
* ~~Pinned Neo dep to 22.04.00a to work-around incompatible neo version~~

```
$ scripts/fetch_data.py -h
usage: Fetches data not included in the repository by default [-h] {fetch,check} ...

optional arguments:
  -h, --help     show this help message and exit

Subcommands:
  valid subcommands

  {fetch,check}
    fetch        Fetch datasets
    check        Check download status of large files. Displays a True/False whether all files are downloaded.
```

Closes #88


BartleyR (approved):

docker/conda/environments/cuda11.4_dev.yml
```diff
@@ -50,7 +50,7 @@ dependencies:
     - isort
     - mlflow>=1.23
     - myst-parser==0.17
-    - neo 22.04.*
+    - neo 22.04.00a*
```
mdemoret-nv: Revert this back to just 22.04

ci/scripts/fetch_data.py
```diff
@@ -0,0 +1,122 @@
...
+    return args
+
+
+def main():
+    args = parse_args()
+    logging.basicConfig(level=logging.INFO, format="%(message)s")
+    include_paths = [LFS_DATASETS[p] for p in args.data_set]
+
+    if os.environ.get('TERM') is not None:
+        curses.wrapper(lfsPull, include_paths)
```
mdemoret-nv: I dont see any output when running this script locally. Which branch should this be taking when run by the user?

mdemoret-nv: I dont see any output when running this script locally. Which branch should this be taking when run by the user?

ci/scripts/fetch_data.py
```diff
@@ -0,0 +1,122 @@
...
+        return output
+
+
+def parse_args():
+    argparser = argparse.ArgumentParser("Fetches data not included in the repository by default")
+    argparser.add_argument("data_set",
+                           nargs='*',
+                           choices=list(LFS_DATASETS.keys()),
+                           help="Data set to fetch")
+    args = argparser.parse_args()
```
mdemoret-nv: I think it would be useful to use subcommands here and add some utility functions that users might need. Something like:

- `fetch`
   - Specify one of `all`, `models', `test`, etc.
- `check`
   - default: shows True/False whether all files are downloaded
   - `--list`: Will list all missing files

Then if we add new options/commands/features in the future, we wont have to have any breaking changes

mdemoret-nv: I think it would be useful to use subcommands here and add some utility functions that users might need. Something like:

- `fetch`
   - Specify one of `all`, `models', `test`, etc.
- `check`
   - default: shows True/False whether all files are downloaded
   - `--list`: Will list all missing files

Then if we add new options/commands/features in the future, we wont have to have any breaking changes

ci/scripts/fetch_data.py
```diff
@@ -0,0 +1,122 @@
...
+                line = progress_lines[-1].decode("UTF8")
+                last_print_len = print_line(stdscr, max_x, last_print_len, line)
+
+            returncode = popen.poll()
+
+        # Check if we have any output
+        output = popen.stdout.read().rstrip("\n")
+
+        if returncode != 0:
+            raise subprocess.CalledProcessError(returncode=returncode, cmd=cmd, output=output)
```
mdemoret-nv: Should probably log the output and error code here before raising an error

mdemoret-nv: Should probably log the output and error code here before raising an error

ci/scripts/fetch_data.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    Performs a git lfs pull.
+    This can take upwards of a minute to complete we will make use of the
+    GIT_LFS_PROGRESS hook to send progress to a file which we can then tail
+    while the command is executing.
+    """
+    cmd = 'git lfs pull -I "{}"'.format(','.join(include_paths))
+    with tempfile.NamedTemporaryFile() as progress_file:
+        env = os.environ.copy()
+        env['GIT_LFS_PROGRESS'] = progress_file.name
```
mdemoret-nv: Have you also looked into trying `GIT_LFS_FORCE_PROGRESS`? This might be an easier way to force the progress output without needing curses.

mdemoret-nv: Have you also looked into trying `GIT_LFS_FORCE_PROGRESS`? This might be an easier way to force the progress output without needing curses.

mdemoret-nv (changes requested):
Couple of comments. I think the script should be moved to `./scripts` filter instead of CI. Also, printing to the console didnt work for me out of the box.

ci/scripts/fetch_data.py
dagardner-nv: Locally it should go down the `curses.wrapper` branch, however when its run on a system that already has all of the lfs assets pulled there isn't any output. 

What I had to do to test this was I had to make `david-lfs-config` the default branch in my fork, and then test this with a fresh clone of my fork.

ci/scripts/fetch_data.py
dagardner-nv: Oh! This is a really good suggestion, I read over the docs but somehow missed this one. The script is much better now, thanks.

ci/scripts/fetch_data.py
dagardner-nv: done

scripts/fetch_data.py
```diff
@@ -0,0 +1,133 @@
...
+        downloaded = parts[1] == '*'
+        if all_downloaded and not downloaded:
+            all_downloaded = False
+
+        if list_all:
+            # the join on 2: is needed to handle file names that contain a blank space
+            logging.info('%s - %s', ' '.join(parts[2:]), downloaded)
+
+    if not list_all:
+        logging.info(all_downloaded)
```
mdemoret-nv, line 95: I ran this and it was not clear at all to me what True/False meant. Can this instead return 0/1 error code (incase this script is used in CI or something) and print something like: "All LFS files downloaded" on true and "Missing the following LFS files: <list of missing files>" on false? Keep the `--list` option the way it currently is.

mdemoret-nv (changes requested):
The check command is good but needs better output.

mdemoret-nv (approved):
Looks good. Great addition

mdemoret-nv:
@gpucibot merge

PR #147, opened 2022-06-08
Documentation fixes

### LINKED ISSUE #92: [DOC] Add NGC references back to README
**Report needed documentation**
We removed the NGC references from the getting started section temporarily as to avoid confusion on the multiple paths to get started. The removed link directed users here: [https://catalog.ngc.nvidia.com/orgs/nvidia/teams/morpheus/collections/morpheus_](https://catalog.ngc.nvidia.com/orgs/nvidia/teams/morpheus/collections/morpheus_). This is where a user would go if they want to use Helm charts to run Morpheus in a K8s environment.

**Describe the documentation you'd like**
This link and wording around it should be added back into the `README.md` file once we can decide on the wording to avoid potential confusion with users mixing and matching paths to get started.

dagardner-nv:
Documentation updates to:

* Reflect Python module re-org, and config changes to Git LFS
* Fixes from Adam Wood
* Remove usage of third-party psutils lib (no longer in Conda env and not strictly needed)
* Remove instructions for building patched cudf
* Update pipeline images for examples

Fixes #92


bsuryadevara (commented):
@dagardner-nv Helm expects unique name for the release. The user must substitute <YOUR RELEASE NAME> with the release name which ever they like. Everything else LGTM

docs/source/morpheus_quickstart_guide.md
```diff
@@ -199,7 +217,7 @@ $ helm fetch https://helm.ngc.nvidia.com/nvidia/morpheus/charts/morpheus-sdk-cli
 $ helm install --set ngc.apiKey="$API_KEY" \
```
bsuryadevara, line 217: Could we move this command to a new section following the "Example Workflows" section, as Adam Wood recommended, add the section title to the table of contents?

bsuryadevara (changes requested):

docs/source/developer_guide/guides/3_simple_cpp_stage.md
```diff
@@ -28,9 +28,9 @@ CppConfig.set_should_use_cpp(False)
 
 If a stage does not have a C++ implementation, Morpheus will fall back to the Python implementation without any additional configuration and operate in a hybrid execution mode.
 
-In addition to C++ accelerated stage implementations, Morpheus also provides a C++ implementation for message primitives. When C++ execution is enabled, constructing one of the Python message classes defined in the `morpheus.pipeline.messages` module will return a Python object with bindings to the underlying C++ implementation.
+In addition to C++ accelerated stage implementations, Morpheus also provides a C++ implementation for message primitives. When C++ execution is enabled, constructing one of the Python message classes defined in the `morpheus.messages` package will return a Python object with bindings to the underlying C++ implementation.
```
drobison00, line 31: morpheus._lib.messages, and its definitely a module, not a package. Morpheus itself is a package, but all its pybind/cython sub modules aren't.

docs/source/developer_guide/guides/3_simple_cpp_stage.md
```diff
@@ -28,9 +28,9 @@ CppConfig.set_should_use_cpp(False)
...
+Since we are defining our pipelines in Python, it becomes the responsibility of the Python implementation to build a C++ accelerated node. This happens in the `_build_source` and `_build_single` methods. Ultimately it is the decision of a Python stage to build a Python node or a C++ node. It is perfectly acceptable to build a Python node when `morpheus.config.CppConfig.get_should_use_cpp()` is configured to `True`. It is not acceptable, however, to build a C++ node when `morpheus.config.CppConfig.get_should_use_cpp() == False`. The reason is the C++ implementations of Morpheus' messages can be consumed by Python and C++ stage implementations alike. However when `morpheus.config.CppConfig.get_should_use_cpp() == False`, the Python implementations of messages will be used which cannot be consumed by the C++ implementations of stages.
```
drobison00, line 33:
```suggestion
Since we are defining our pipelines in Python, it becomes the responsibility of the Python implementation to build a C++ accelerated node. This happens in the `_build_source` and `_build_single` methods. Ultimately it is the decision of a Python stage to build a Python node or a C++ node. It is perfectly acceptable to build a Python node when `morpheus.config.CppConfig.get_should_use_cpp()` is configured to `True`. It is not acceptable, however, to build a C++ node when `morpheus.config.CppConfig.get_should_use_cpp() == False`. The reason is the C++ implementations of Morpheus' messages can be consumed by Python and C++ stage implementations alike. However when `morpheus.config.CppConfig.get_should_use_cpp() == False`, the Python implementations of messages will be used which cannot be consumed by the C++ implementations of stages.
```
```suggestion
Since we are defining our pipelines in Python, it becomes the responsibility of the Python implementation to build a C++ accelerated node. This happens in the `_build_source` and `_build_single` methods. Ultimately it is the decision of a Python stage to build a Python node or a C++ node. It is perfectly acceptable to build a Python node when `morpheus.config.CppConfig.get_should_use_cpp()` is configured to `True`. It is not acceptable, however, to build a C++ node when `morpheus.config.CppConfig.get_should_use_cpp() == False`. The reason is the C++ implementations of Morpheus' messages can be consumed by Python and C++ stage implementations alike. However when `morpheus.config.CppConfig.get_should_use_cpp() == False`, the Python implementations of each message type will be used which cannot be consumed by the C++ implementations of stages.
```

docs/source/index.rst
```diff
@@ -41,7 +41,7 @@ Features
  * Real-Time Telemetry
     * Morpheus can receive rich, real-time network telemetry from every NVIDIA® BlueField® DPU-accelerated server in the
       data center without impacting performance. Integrating the framework into a third-party cybersecurity offering
-      brings the world’s best AI computing to communication networks.
+      brings the world's best AI computing to communication networks.
```
drobison00, line 44: Good catch.

docs/source/index.rst
```diff
@@ -100,6 +100,65 @@ Launch an interactive container to start using Morpheus:
...
 See :doc:`basics/overview` for more information on using the CLI.
 
+Building Localy
+---------------
+To get started, first clone the Morpheus repo:
+
+.. code-block:: bash
+
+   export PYTHON_VER=3.8
+   export RAPIDS_VER=22.04
```
drobison00, line 110: Probably should be 22.06 based on @cwharris 's upcoming changes.

morpheus/utils/type_utils.py
```diff
@@ -116,6 +116,9 @@ def pretty_print_type_name(t: typing.Type) -> str:
     Determines a good label to use for a type. Keeps the strings shorter.
     """
 
+    if t is None:
+        return ""
+
```
drobison00, line 121:
```suggestion
    if t is None:
        return "NoneType"

```

Not sure if there is a reason to return an empty string, but something more descriptive seems like a good idea if its not problematic.

drobison00 (changes requested):

docs/source/index.rst
dagardner-nv: Thanks

morpheus/utils/type_utils.py
dagardner-nv: I added this bit to get the `Pipeline.visualize` method working for the sole purpose of updating the pipeline graphs to reflect the current state of the examples (ie. dropping the Buffer & Delay stages).

It looks like in the past this displayed the message type between stages, so I just hacked this bit in. Returning a "NoneType" string would cause that to be printed between stages.

drobison00: Gotcha. Well, my vote would be to have something between nodes rather than nothing.

docs/source/index.rst
dagardner-nv: I checked, and that PR (#180) is targeting branch-22.08.
@cwharris do you know if moving to cudf=22.06 is going to be merged back into the 22.06 release?

morpheus/utils/type_utils.py
dagardner-nv: Maybe a "->"?
Otherwise I could probably just spend a little more time debugging the visualize code, my guess is it just got broken when the Messages were refactored

docs/source/index.rst
cwharris: I had it targeting the wrong branch but it's merged in 22.06 now.

morpheus/utils/type_utils.py
dagardner-nv: Split this one off into its own issue #188

docs/source/index.rst
dagardner-nv: Updated, @drobison00 & @cwharris thanks

morpheus/utils/type_utils.py
dagardner-nv: I reverted this change. The ultimate source of the problem is that the Morpheus pipeline doesn't have access to the type information of the messages between nodes until the pipeline is built, which happens asynchronously, without a way to wait for completion. 

The visualize method gets called early before the pipeline starts so its a race condition between it and the completion of the of the pipeline build. When this happens the `visualize` method gets values of `None`

dagardner-nv:
I updated the pipeline graphs to include message types

mdemoret-nv (changes requested):
David, I saw you started to add the NGC references to `index.rst` which is aligned with what needs to happen to close out #92. Can you do one more pass over the repo with the changes requested in #92 so we can close that out as well with this PR?

To add a little more info, the `README.md` and `CONTRIBUTING.md` in the repo will be seen by developers and users before a release has happened. So there shoundnt be any references to NGC in these files and there shouldnt be any references to the current release (i.e. we want to avoid statement like `pull nvcr.io/morpheus:v22.08` since 22.08 hasnt even shipped yet)

The corollary to this is the documentation in `docs/*.rst` **will** be seen after a release has happened. And the user might be viewing these docs on docs.nvidia.com. So in these files, we dont want to make the assumption that the user has the repo cloned. So all references to things like `./docker/run_container_release.sh` should be removed in favor of the `docker run ...` command. You already started doing some of this.

Let me know if you have any questions

mdemoret-nv (changes requested):
Im still seeing references to NGC in the root `README.md`

mdemoret-nv (approved):
I updated the README with my last comments. Looks good now.

drobison00 (approved):

mdemoret-nv:
@gpucibot merge

PR #181, opened 2022-06-23
GPU not needed for build & documentation CI stages

### LINKED ISSUE #115: [FEA] Documentation build stage shouldn't require a gpu
If we try to build without it we get this error:
```
F20220518 00:39:57.558239   679 device_info.cpp:39] Check failed: nvmlInit_v2() == NVML_SUCCESS (9 vs. 0) Failed to initialize NVML
```

Sphinx provides the ability to mock modules during the build phase, likely we just need to identify these and add them to our config.

dagardner-nv:
Also simplifies build script.

Fixes #115

### DIFF
--- a/ci/Jenkinsfile
+++ b/ci/Jenkinsfile
@@ -114,14 +114,14 @@ pipeline {
                     }
                     agent {
                         docker {
-                            image 'gpuci/rapidsai:22.04-cuda11.5-devel-centos7-py3.8'
-                            label 'driver-495'
-                            args '--runtime "nvidia" -e "NVIDIA_VISIBLE_DEVICES=$EXECUTOR_NUMBER"'
+                            image 'gpuci/rapidsai-driver:22.04-cuda11.5-devel-centos7-py3.8'
+                            label 'cpu'
                         }
                     }
                     environment {
                         HOME = "${WORKSPACE}"
                         MORPHEUS_NO_CPP = '1'
+                        PARALLEL_LEVEL = '4'
                     }
                     steps {
                         cleanWs(
--- a/ci/Jenkinsfile
+++ b/ci/Jenkinsfile
@@ -44,15 +44,14 @@ pipeline {
                         timeout(time: 4, unit: 'HOURS')
                     }
                     environment {
-                        PARALLEL_LEVEL = '10'
+                        PARALLEL_LEVEL = '16'
                         USE_SCCACHE = '1'
                         HOME = "${WORKSPACE}"
                     }
                     agent {
                         docker {
-                            image 'gpuci/rapidsai:22.04-cuda11.5-devel-centos7-py3.8'
-                            label 'driver-495'
-                            args '--runtime "nvidia" -e "NVIDIA_VISIBLE_DEVICES=$EXECUTOR_NUMBER"'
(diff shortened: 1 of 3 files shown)


drobison00 (approved):
LGTM

ci/scripts/jenkins/test.sh
```diff
@@ -75,5 +75,3 @@ tar cfj ${WORKSPACE_TMP}/test_reports.tar.bz $(basename ${REPORTS_DIR})
 
 gpuci_logger "Pushing results to ${DISPLAY_ARTIFACT_URL}"
 aws s3 cp ${WORKSPACE_TMP}/test_reports.tar.bz "${ARTIFACT_URL}/test_reports.tar.bz"
-
-exit ${TEST_RESULTS}
```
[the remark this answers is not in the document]
dagardner-nv: Ahh you caught me in the middle of trying something out, 
I reverted this.

ci/Jenkinsfile
```diff
@@ -44,15 +44,14 @@ pipeline {
                         timeout(time: 4, unit: 'HOURS')
                     }
                     environment {
-                        PARALLEL_LEVEL = '10'
+                        PARALLEL_LEVEL = '16'
```
mdemoret-nv, line 47: Is this always going to be 16? Is there no environment variable set by the Jenkins runtime? How do they do it in RMM?

ci/scripts/jenkins/build.sh
```diff
@@ -23,28 +23,20 @@ rm -rf ${MORPHEUS_ROOT}/.cache/ ${MORPHEUS_ROOT}/build/
...
-show_conda_info
-
-gpuci_logger "Installing dependencies"
-mamba env update -q -n morpheus -f ${MORPHEUS_ROOT}/docker/conda/environments/cuda${CUDA_VER}_dev.yml
-conda deactivate && conda activate morpheus
-
-gpuci_logger "Final Conda Environment"
-conda list
-
-gpuci_logger "Check cmake & ninja"
```
mdemoret-nv, line 47: Lets keep this log

mdemoret-nv (approved):
Approving since my comments are very minor. Once the conversations are resolved, this should be good to go.

ci/scripts/jenkins/build.sh
dagardner-nv: I dropped the call to `conda list` because a few lines prior the script is calling the `show_conda_info` function which displays the same info.

I also dropped the "Check cmake & ninja" bit but still log the cmake and ninja versions. 

Mostly this is because in the past we build the conda env in three stages:
1) The stuff needed to build cudf
2) cudf
3) everything else

but now all that stuff went away

ci/Jenkinsfile
mdemoret-nv: I think that would be fine.

ci/scripts/jenkins/build.sh
mdemoret-nv: I guess my comment should have been just: Lets keep this single line: `gpuci_logger "Check cmake & ninja"`. I just wanted the header so its easier to break up the log into sections.

mdemoret-nv: I see all versions are reported in one block

mdemoret-nv:
@gpucibot merge

PR #164, opened 2022-06-15
release container build fixes

cwharris:
Changes required to build the release container.

Closes #166

PR #200, opened 2022-06-27
Update Dockerfile for label data dir

pdmack:
Make sure we keep the label, column data dir for the existing pipelines.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -146,6 +146,7 @@ COPY "./docker" "./docker"
 COPY "./docs" "./docs"
 COPY "./examples" "./examples"
 COPY "./models" "./models"
+COPY "./morpheus/data" "./morpheus/data"
 COPY "./scripts" "./scripts"
 COPY ["*.md", "LICENSE", "./"]
 


docker/Dockerfile
```diff
@@ -146,6 +146,7 @@ COPY "./docker" "./docker"
 COPY "./docs" "./docs"
 COPY "./examples" "./examples"
 COPY "./models" "./models"
+COPY "./morpheus/data" "./morpheus/data"
```
mdemoret-nv, line 149: This directory gets installed now with the `morpheus` python package. If you run:

```bash
$ python -c "import morpheus;print(morpheus.DATA_DIR)"
$HOME/Repos/morpheus/morpheus-dev/morpheus/data
```
Is this path used by scripts anywhere?

mdemoret-nv (commented):
The data path is now installed as part of the python/conda package. Does this PR need to be more about documentaiton/scripting updates?

docker/Dockerfile
pdmack: Let me double check things once we have a working release build. I'll validate that our external instructions are correct.

mdemoret-nv: Release build should be working as of #164

mdemoret-nv:
Closing this in favor of #232.

PR #130, opened 2022-05-27
Split data dir, moving large files into examples/data

dagardner-nv:
Brings directory size down to 1.6MB down from 270MB

+ Move large files to `exaples/data` 
+ Migrate existing files in `exaples/data` to git-lfs
+ Add missing `email_with_addresses.jsonlines` needed for phishing detection developer guide
+ Update file paths in docs for data files
+ Update import paths in developer guide

Depends on changes in ~~#62~~
Fixes #120

PR #232, opened 2022-06-28
Add CLI Relative Path Fallback

mdemoret-nv:
In PR #130 we moved the `data` directory into `morpheus/data` and installed it with the python package. This required changing some of the default CLI arguments from relative paths like `data/labels_nlp.txt` to absolute paths like `morpheus.DATA_DIR/labels_nlp.txt`.

To make it easy for the user to see how to change the labels file, we respecified the default argument value in documentation (i.e. `--labels_file=data/labels_nlp.txt`). Now that this needs to be an absolute path, the command in the documentation does not work. Adding absolute paths in the documentation is not feasible since this would require very long paths that would change from machine to machine.

Instead, if the user specifies a data file with a relative path, we first check to see if a file exists relative to the current working directory. If it doesnt exist, then we check for a relative file to the current morpheus install. This allows commands from the documentation like: `morpheus run pipeline-nlp --labels_file=data/labels_nlp.txt` to find the correct path. We only choose the fallback value when no other file is found.

Related to PR #200

### DIFF
--- a/morpheus/cli.py
+++ b/morpheus/cli.py
@@ -14,6 +14,7 @@
 
 import logging
 import os
+import typing
 import warnings
 from functools import update_wrapper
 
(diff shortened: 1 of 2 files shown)


dagardner-nv (approved):

mdemoret-nv:
Merging this to test with release packages tomorrow.

mdemoret-nv:
@gpucibot merge

PR #234, opened 2022-06-28
Revert ransomware feature config changes

bsuryadevara:
Reverted ransomware feature config changes as dask is overriding config values.

### DIFF
--- a/examples/ransomware_detection/stages/create_features.py
+++ b/examples/ransomware_detection/stages/create_features.py
@@ -59,10 +59,9 @@ class CreateFeaturesRWStage(MultiMessageStage):
         threads_per_worker: int = 2,
     ):
         self._client = Client(threads_per_worker=threads_per_worker, n_workers=n_workers)
+        self._feature_config = FeatureConfig(file_extns, interested_plugins)
+        self._feas_all_zeros = dict.fromkeys(feature_columns, 0)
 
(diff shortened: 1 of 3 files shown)


examples/ransomware_detection/stages/create_features.py
```diff
@@ -59,10 +59,9 @@ def __init__(
         threads_per_worker: int = 2,
     ):
         self._client = Client(threads_per_worker=threads_per_worker, n_workers=n_workers)
+        self._feature_config = FeatureConfig(file_extns, interested_plugins)
+        self._feas_all_zeros = dict.fromkeys(feature_columns, 0)
```
mdemoret-nv, line 63: Is `self._feas_all_zeros` used anywhere else? Do we need to save it in `self`?

mdemoret-nv (approved):
Just one question. I like this better than saving `_config`

examples/ransomware_detection/stages/create_features.py
bsuryadevara: using it here: https://github.com/nv-morpheus/Morpheus/blob/7b0bb90431921041954d4b5ebe75d23583110eeb/examples/ransomware_detection/stages/create_features.py#L114

mdemoret-nv:
@gpucibot merge

PR #257, opened 2022-07-01
Fixing the from-kafka stage

mdemoret-nv:
Fixes #254 

This re-enables the check on the number of messages received before trying to process to cudf.

PR #262, opened 2022-07-05
Fixing Python Kafka Source with Multiple Threads

mdemoret-nv:
A small regression in PR #257 that would duplicate messages in the `from-kafka` stage when `--use_cpp=False`. If you used 3 threads you would process each message in Kafka 3 times instead of dividing the messages between threads.

This also improves the shutdown behavior of the `from-kafka` stage when pressing Ctrl+C to close the pipeline.


dagardner-nv (approved):

mdemoret-nv:
@gpucibot merge

PR #273, opened 2022-07-13
Ensure TableInfo only appears in one lib

### LINKED ISSUE #271: [BUG] table_info appears in two different libs
**Describe the bug**
`morpheus/_lib/src/objects/table_info.cpp` appears in both `morpheus/_lib/cmake/libraries/cuda_utils.cmake` and `morpheus/_lib/cmake/libraries/morpheus.cmake`.

Which places it in both `libmorpheus_utils.so` and `libmorpheus.so`.
Not sure if this is currently causing any issues, but likely just a copy/paste error.

build morpheus

**Expected behavior**
Likely it only needs to be in `libmorpheus_utils.so`.

dagardner-nv:
* Removes `table_info.cpp` from `morpheus.cmake`
* Sets visibility on the `TableInfo.as_py_object` method, this was showing up as undefined in `libmorpheus.so`. 

I tried just setting `#pragma GCC visibility push(default)` at the top of the class definition, but this didn't work, according to the gcc docs this is intentional:
> In C++, ‘#pragma GCC visibility’ affects only namespace-scope declarations. Class members and template specializations are not affected; if you want to override the visibility for a particular member or instantiation, you must use an attribute.
https://gcc.gnu.org/onlinedocs/gcc-4.9.2/gcc/Visibility-Pragmas.html

fixes #271

### DIFF
--- a/morpheus/_lib/cmake/libraries/cuda_utils.cmake
+++ b/morpheus/_lib/cmake/libraries/cuda_utils.cmake
@@ -12,6 +12,10 @@
 # the License.
 # =============================================================================
 
+list(APPEND CMAKE_MESSAGE_CONTEXT "cuda_utils")
+
+find_package(pybind11 REQUIRED)
+
 add_library(cuda_utils
     SHARED
       ${MORPHEUS_LIB_ROOT}/src/objects/dev_mem_info.cpp
@@ -34,6 +38,7 @@ target_link_libraries(cuda_utils
       matx::matx
       cudf::cudf
       Python3::NumPy
+      pybind11::pybind11
 )
 
 set_target_properties(cuda_utils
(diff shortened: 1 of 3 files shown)


drobison00 (approved):
Looks reasonable.

pdmack:
LGTM, no failures from a file smoke test
```
(morpheus) root@sdk-cli-sdk1:/workspace# nm -C /opt/conda/envs/morpheus/lib/python3.8/site-packages/morpheus/_lib/libmorpheus.so | grep table_info.cpp
(morpheus) root@sdk-cli-sdk1:/workspace# nm -C /opt/conda/envs/morpheus/lib/python3.8/site-packages/morpheus/_lib/libmorpheus_utils.so | grep table_info.cpp
00000000000332f0 t _GLOBAL__sub_I_table_info.cpp
```

mdemoret-nv (approved):
LGTM

mdemoret-nv:
@gpucibot merge

PR #283, opened 2022-07-18
Update GNN example readme

pdmack:
`chardet` package now required by stellargraph
Also, switching to mamba for speed and conda seemed to struggle solving env.
Not sure if we should start pinning these?


cwharris (approved):
LGTM

PR #300, opened 2022-07-22
Simplify Python impl for KafkaSourceStage

### LINKED ISSUE #294: [BUG] KafkaSourceStage Python & C++ source impls commit differently
**Describe the bug**
The Python impl for the `KafkaSourceStage` commits asynchronously when the Morpheus message constructed from the payload of the Kafka message is garbage collected:
```python
# Once the meta goes out of scope, commit it
def commit(topic, part_no, keys, lowest, offset):
...
if (not self._disable_commit):
    weakref.finalize(meta, commit, *part[1:])
```

Since `MessageMeta` is typically preserved throughout the entire pipeline, this has the effect of not committing until `meta` has been successfully processed by the sink. This has the benefit that any messages in-flight when a pipeline is shut-down are not committed, and will be re-processed on a restart.

However this has the down-side of not being compatible with any downstream stages which make use of `ops.to_list`, this is probably OK as only the validation and trigger stages do this. This also makes assumptions about when the message is garbage collected, and assumes that the underlying Kafka consumer hasn't yet been garbage collected and still has a valid connection to Kafka.

The C++ impl however performs a commit after processing the messages:
```cpp
auto should_commit = rebalancer.process_messages(message_batch);

if (should_commit)
{
    CHECK_KAFKA(consumer->commitAsync(), RdKafka::ERR_NO_ERROR, "Error during commitAsync");
}
```

This has the negative side-effect that any in-flight messages could be lost during an unexpected shutdown of the pipeline.

Execute a pipeline utilizing the `KafkaSourceStage`

**Expected behavior**
The behavior of the Python impl is preferred

**Additional context**
Can we attach some sort of callback to `MessageMeta` which can then be called by the destructor? 

Maybe the callback could explicitly invoked by the sink allowing the commit to occur consistently between Python & C++ impls, and avoids depending on the behavior of the Python gc.

dagardner-nv:
Current impl is overly complicated making use of undocumented methods.
Removes dependency on `cudf_kafka`.

Fixes #294


mdemoret-nv (changes requested):
Are there any tests that compare the C++ and Python implementations?

dagardner-nv:
> Are there any tests that compare the C++ and Python implementations?

I just added one, its a bit tricky as commits perform asynchronously. I had to add a new bool constructor argument to enable blocking commits. 

Even then the C++ impl commits just after calling `on_next`

morpheus/stages/input/kafka_source_stage.py
```diff
@@ -147,221 +134,93 @@ def stop(self):
...
-
-                        if high > lowest + self._max_batch_size:
-                            high = lowest + self._max_batch_size
-                        if high > lowest:
-                            out.append((consumer_params, self._topic, partition, self._keys, lowest, high - 1))
-                            positions[partition] = high
-
-                    consumer_params['auto.offset.reset'] = 'earliest'
+    def _source_generator(self):
+        # TODO : Needs to batch records until _stop_requested, _PARTITION_EOF or batch size has been hit
```
mdemoret-nv, line 173: Is this `TODO` still needed?

mdemoret-nv (approved):
Looks good. Lets get `branch-22.11` merged so we can get CI passing again.

morpheus/stages/input/kafka_source_stage.py
dagardner-nv: Good catch

mdemoret-nv:
@gpucibot merge

PR #312, opened 2022-08-01
Add CLI Extensions to Allow Running Custom Stages with the Morpheus CLI

### LINKED ISSUE #85: [FEA] Make the Morpheus CLI Extendable with the Ability to Add New Commands
All pipelines that define at least 1 custom Stage must use the Python interface and cannot use the CLI. This is because there is no way to register new stages with the CLI unless they are built as first-party stages. This forces all of our built-in examples to use the python interface, duplicating a lot of click commands.

Custom stages could be registered as commands with the CLI using various methods (such as a decorator). Once registered, CLI commands simply forward all options as `kwargs` to the constructor of each stage.

**Additional context**
Custom stages could be automatically discovered using the `entry_point` functionality in python. This is used by tools such as `pytest` to build extensions into the pytest CLI.

mdemoret-nv:
This PR allows users to run pipelines via the CLI, even if they use custom built stages. This works by adding 2 new capabilities: A plugin system for the CLI, and a new `@register_stage()` decorator.

The plugin system utilizes the same functionality as `pytest` (via the `pluggy` library) and adds 2 hooks into the CLI:

- `morpheus_cli_collect_stage_names`
   - Returns a list of stage names for the supplied pipeline mode
- `morpheus_cli_make_stage_command`
   - Returns a `click.Command` given the supplied pipeline mode and name

The new `@register_stage` decorator allows users to register a stage with the CLI. The arguments, name, and help strings are all pulled from the stage's type annotations and docstrings. For example, one can register a stage with the following:
```python
@register_stage("from-cloudtrail", modes=[PipelineModes.AE])
class CloudTrailSourceStage(SingleOutputSource):
    ...
```

While this works well now, there are some outstanding issues to address:

- [x] The plugin system and decorator are somewhat mutually exclusive
   - [x] Its entirely possible to use one and not the other
- [x] Our own stage's do not use this new functionality
   - [x] The main reason is that each stage needs to be imported to get the command information. When using autocomplete, this takes a very long time to load all of the necessary libraries, most of which will not be used.

Closes #85


dagardner-nv (approved):

mdemoret-nv:
@gpucibot merge

PR #316, opened 2022-08-04
Fixes issues with NLP pipelines when data is not truncated

### LINKED ISSUE #305: [BUG] SID Crash with `--use_cpp=True` and `--truncation=False`
**Describe the bug**
When running the SID workflow with `--use_cpp=True` and `--truncation=False`, the pipeline will crash when trying to slice the inference message with the error: 
```
F20220725 20:03:53.547375  6967 multi_inference.cpp:80] Check failed: this->mess_count == this->count At this time, mess_count and count must be the same for slicing
```

The following launch command should produce the bug:
```bash
morpheus --log_level=DEBUG run --num_threads=1 --pipeline_batch_size=1024 --model_max_batch_size=32 --edge_buffer_size=4 --use_cpp=True \
   pipeline-nlp --model_seq_length=256 \
      from-file --filename=models/datasets/validation-data/sid-validation-data.csv --repeat=5 \
      deserialize \
      preprocess --vocab_hash_file=./morpheus/data/bert-base-uncased-hash.txt --truncation=False --do_lower_case=True --add_special_tokens=False \
      inf-triton --model_name=sid-minibert-onnx --server_url=localhost:8001 --force_convert_inputs=True \
      monitor --description Inference\ Rate --unit inf \
      add-class \
      serialize --exclude \^_ts_ \
      to-file --filename=./.tmp/detections2-nlp.csv --overwrite
```

**Expected behavior**
The pipeline should run to completion with the same results as the `--use_cpp=False` optino.

**Additional context**
The warning message is likely because the implementation of `TensorObject::read_element` is not correct. At first glance, the implementation of that function should look something like:
```cpp
template <typename T, size_t N>
    T read_element(const TensorIndex (&idx)[N]) const
    {
        auto stride = this->get_stride();
        auto shape  = this->get_shape();

        CHECK(std::transform_reduce(
            shape.begin(), shape.end(), std::begin(idx), 0, std::logical_and<>(), std::greater<>()))
            << "Index is outsize of the bounds of the tensor. Index="
            << detail::array_to_str(std::begin(idx), std::begin(idx) + N)
            << ", Size=" << detail::array_to_str(shape.begin(), shape.end()) << "";

        CHECK(DataType::create<T>() == this->dtype())
            << "read_element type must match array type. read_element type: '" << DataType::create<T>().name()
            << "', array type: '" << this->dtype().name() << "'";

        size_t offset = std::transform_reduce(
                            stride.begin(), stride.end(), std::begin(idx), 0, std::plus<>(), std::multiplies<>()) *
                        this->dtype_size();

        T output;

        SRF_CHECK_CUDA(
            cudaMemcpy(&output, static_cast<uint8_t*>(this->data()) + offset, sizeof(T), cudaMemcpyDeviceToHost));

        return output;
    }
```

dagardner-nv:
* Fixes type-os in `TensorObject::read_element` as suggested in description in #305
* In cases where the length of the output results do not match the length of the data frame the `seq_ids` array is used to perform reduction of the rows. Such that if rows 5,6 & 7 of the output results map to row 5 in the dataframe, the max value of each row is stored in the response output.
* Add new method `MatxUtil::reduce_max` to perform reduction.

fixes #305


dagardner-nv:
rerun tests

dagardner-nv:
rerun tests

morpheus/_lib/include/morpheus/objects/tensor_object.hpp
```diff
@@ -476,8 +476,10 @@ struct TensorObject final
         auto stride = this->get_stride();
         auto shape  = this->get_shape();
 
+        DCHECK(shape.size() == N) << "Length of idx must match lengh of shape";
```
mdemoret-nv, line 479: I think this should be a `CHECK`

morpheus/_lib/include/morpheus/objects/tensor_object.hpp
```diff
@@ -504,8 +506,10 @@ struct TensorObject final
         auto stride = this->get_stride();
         auto shape  = this->get_shape();
 
-        CHECK(std::transform_reduce(
-            stride.begin(), stride.end(), std::begin(idx), 0, std::logical_and<>(), std::less<>()))
+        DCHECK(shape.size() == N) << "Length of idx must match lengh of shape";
```
mdemoret-nv, line 509: I think this should be a `CHECK`

mdemoret-nv (approved):
Just 2 small changes

dagardner-nv:
rerun tests

mdemoret-nv:
@gpucibot merge

PR #360, opened 2022-09-16
Removing `grpcio-channelz`

mdemoret-nv:
Looks like the `grpcio-channelz` package has been breaking CI. Since this package isnt even used anywhere since the `streamz` days, we should remove it.


dagardner-nv (approved):
Gladly approving

mdemoret-nv:
@gpucibot merge

PR #372, opened 2022-09-23
Use libcudacxx from NVIDIA repo

pdmack:
Remove @ryanolson private fork of libcudacxx now that (we think) the required memory manager is in the upstream.

Rapids latest is 1.8.0
rapidsai/rapids-cmake#253


cmake/dependencies.cmake
```diff
@@ -19,7 +19,6 @@ list(APPEND CMAKE_MESSAGE_CONTEXT "dep")
 set(CPM_SOURCE_CACHE "${CMAKE_SOURCE_DIR}/.cache/cpm")
 # Prevent cpm_init from trying to tell us where to put cpm.cmake
 include(get_cpm)
-rapids_cpm_init(OVERRIDE "${CMAKE_CURRENT_SOURCE_DIR}/cmake/deps/rapids_cpm_package_overrides.json")
```
cwharris, line 22: can we also remove json file?

cwharris (approved):

PR #398, opened 2022-10-11
Update Quickstart Guide

bsuryadevara:
- To make the command easier to copy and paste, the `$` symbol was removed.
- Removed duplicate table of contents


bsuryadevara:
> Can you look over the entire repo and make sure this is consistent everywhere? The rule should be:
> 
> * If there is no output from the command shown, any leading `$` or `#` should be removed so the command can be copied
> * If the command is followed by output, a leading `$ ` should be added to differentiate the command from the output

@mdemoret-nv removed leading `$` at all the places as per the instruction.

BartleyR:
Are we good to merge this then?

bsuryadevara:
> @bsuryadevara @mdemoret-nv Are we good to merge this then?

There are no other locations with leading `$`, so merging should be okay.

mdemoret-nv:
rerun tests

mdemoret-nv:
@gpucibot merge

PR #403, opened 2022-10-19
[DRAFT] update dfp starter example

bsuryadevara:
- Updated DFP starter example to use production stages for cloudtrail logs


mdemoret-nv:
Closing these changes since we will be deprecating the starter example instead of unifying them in the future.

PR #406, opened 2022-10-19
Updating to use C++20

mdemoret-nv:
This updates the Morpheus repo to follow SRF and upgrade to C++20. Changes are minor. This does include an additional change to switch the `CMAKE_BUILD_RPATH_USE_ORIGIN` back to `OFF`. This was enabled by default to allow for copying the build folder in CI. However, it breaks local builds that want to use a custom branch of SRF. Instead it has been disabled by default and enabled just in CI.

SRF PR: https://github.com/nv-morpheus/SRF/pull/156


ci/conda/recipes/python-dbg/conda_build_config.yaml
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 c_compiler_version:
-  - 9.4
+  - 11.2
 
 cxx_compiler_version:
-  - 9.4
+  - 11.2
```
drobison00, line 20: Definitely like this; 9.4 is getting pretty old.

mdemoret-nv:
All changes were incorporated in #530. Closing

PR #416, opened 2022-10-25
Pin camouflage to v0.9 and ensure pytest-benchmark>=4

dagardner-nv:
v0.10 was released on 2022-10-25 and completely breaks our tests which depends on it.
pytest-benchmark 4.0 fixes an incompatibility with pytest 7.2

### DIFF
--- a/docker/conda/environments/requirements.txt
+++ b/docker/conda/environments/requirements.txt
@@ -6,6 +6,7 @@
 # Packages listed here should also be listed in setup.py
 git+https://github.com/nv-morpheus/dfencoder.git@branch-22.09#egg=dfencoder
 nvidia-pyindex
+pytest-benchmark>=4.0
 torch==1.10.2+cu113
 tritonclient[all]==2.17.*
 websockets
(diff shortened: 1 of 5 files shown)


docker/conda/environments/requirements.txt
```diff
@@ -6,6 +6,7 @@
 # Packages listed here should also be listed in setup.py
 git+https://github.com/nv-morpheus/dfencoder.git@branch-22.09#egg=dfencoder
 nvidia-pyindex
+pytest-benchmark>=4.0
```
mdemoret-nv, line 9: Why is this here instead of in the conda environment?

mdemoret-nv (commented):
Quick question

docker/conda/environments/requirements.txt
dagardner-nv: I originally added it in the yaml, but conda would report that nothing provides the package. Putting it in the requirements worked.

dagardner-nv: At the time yesterday pytest-benchmark 4.0 hadn't been uploaded to conda-forge, it is now, so I could move it back to the yaml.

mdemoret-nv (approved):
LGTM

dagardner-nv:
@gpucibot merge

PR #427, opened 2022-10-31
Add Locking to `TableInfo` to Support Mutating a DataFrame from Multiple Threads

mdemoret-nv:
If you run multiple threads with the `AddScores` or `AddClassification` stages, they may try to append a new column to the dataframe. In this situation, you can have race conditions where some threads are using the dataframe object at the same time you are adding/removing columns.

To fix this, this PR separates the `TableInfo` class into two classes:

1. `TableInfo`
   2. Allow for read operations to happen in parallel
3. `MutableTableInfo`
   4. Requires exclusive access to the underlying dataframe before it can be constructed. Allows for add/remove operations on columns

The effect is that holding an instance of `TableInfo` will ensure that the column types and count do not change for the lifetime of that object. Holding an instance of `MutableTableInfo` ensures that you have exclusive access to the dataframe and can mutate it.

Outstanding questions:

- [ ] How do we handle the `as_py_object()` method on `TableInfo`? This should probably make a copy.
- [ ] How do we properly test concurrent access?


morpheus/_lib/src/messages/multi.cpp
```diff
@@ -43,6 +44,40 @@
 #include <vector>
 // IWYU pragma: no_include <unordered_map>
 
+namespace {
+
+struct DispatchedCopy
```
dagardner-nv, line 49: This shouldn't be here anymore

mdemoret-nv: Good catch.

mdemoret-nv:
rerun tests

mdemoret-nv:
Moving to 23.01 since the majority of the multi-threaded issues have been addressed.

morpheus/_lib/src/messages/meta.cpp
```diff
@@ -157,15 +158,35 @@ cudf::size_type MessageMetaInterfaceProxy::count(MessageMeta &self)
...
-    // self.m_pydf;.metadata); py::object py_table  =
-    // py::reinterpret_steal<py::object>((PyObject*)py_table_struct);
-
-    // // py_col.inc_ref();
+    // Release any GIL
+    pybind11::gil_scoped_release no_gil;
 
     // return py_table;
-    return self.get_py_table();
+    return self.get_info().copy_to_py_object();
```
dagardner-nv, line 165: If we make this change, we should also make a copy in the Python impl of this class.

Otherwise we will get different behaviors in Python stages when C++ execution is enabled/disabled.

mdemoret-nv:
Closing in favor of #586

PR #428, opened 2022-11-01
426 bug msg keyerror data

bsuryadevara:
- Users can pass the name of a column that has to be pre-processed more easily with parameterized pre-process columns.

### DIFF
--- a/morpheus/_lib/include/morpheus/stages/preprocess_nlp.hpp
+++ b/morpheus/_lib/include/morpheus/stages/preprocess_nlp.hpp
@@ -54,6 +54,7 @@ class PreprocessNLPStage
                        bool truncation,
                        bool do_lower_case,
                        bool add_special_token,
+                       std::string column,
                        int stride = -1);
 
   private:
@@ -63,6 +64,7 @@ class PreprocessNLPStage
     subscribe_fn_t build_operator();
 
     std::string m_vocab_hash_file;
+    std::string m_column;
     uint32_t m_sequence_length;
     bool m_truncation;
     bool m_do_lower_case;
@@ -86,6 +88,7 @@ struct PreprocessNLPStageInterfaceProxy
                                                                           bool truncation,
                                                                           bool do_lower_case,
                                                                           bool add_special_token,
+                                                                          const std::string column
                                                                           int stride = -1);
 };
 #pragma GCC visibility pop
--- a/morpheus/_lib/include/morpheus/stages/preprocess_nlp.hpp
+++ b/morpheus/_lib/include/morpheus/stages/preprocess_nlp.hpp
@@ -88,7 +88,7 @@ struct PreprocessNLPStageInterfaceProxy
                                                                           bool truncation,
                                                                           bool do_lower_case,
                                                                           bool add_special_token,
-                                                                          const std::string column,
+                                                                          const std::string &column,
                                                                           int stride = -1);
 };
 #pragma GCC visibility pop
--- a/morpheus/_lib/include/morpheus/stages/preprocess_nlp.hpp
+++ b/morpheus/_lib/include/morpheus/stages/preprocess_nlp.hpp
@@ -54,7 +54,7 @@ class PreprocessNLPStage
                        bool truncation,
                        bool do_lower_case,
                        bool add_special_token,
-                       int stride = -1,
+                       int stride         = -1,
                        std::string column = "data");
 
   private:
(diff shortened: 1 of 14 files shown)


cwharris:
rerun tests

cwharris (approved):

raykallen (approved):

morpheus/stages/preprocess/preprocess_nlp_stage.py
```diff
@@ -63,6 +63,8 @@ class PreprocessNLPStage(PreprocessBaseStage):
         overflowing token-ids can contain duplicated token-ids from the main sequence. If max_length is equal to stride
         there are no duplicated-id tokens. If stride is 80% of max_length, 20% of the first sequence will be repeated on
         the second sequence and so on until the entire sentence is encoded.
+    column : str
+        Name of the column containing the data that needs to be preprocessed.
```
mdemoret-nv, line 67:
```suggestion
    column : str, default = "data"
        Name of the column containing the data that needs to be preprocessed.
```

mdemoret-nv (changes requested):
We need a test case to exercise this new functionality.

morpheus/_lib/include/morpheus/stages/preprocess_nlp.hpp
```diff
@@ -86,6 +88,7 @@ struct PreprocessNLPStageInterfaceProxy
                                                                           bool truncation,
                                                                           bool do_lower_case,
                                                                           bool add_special_token,
+                                                                          const std::string column,
```
dagardner-nv, line 91: I would make this a 
```suggestion
                                                                          const std::string& column,
```

dagardner-nv (approved):
LGTM 

minor nitpick to make one of the arguments a reference, but looks good in general.

morpheus/stages/preprocess/preprocess_nlp_stage.py
bsuryadevara: @mdemoret-nv I have updated the existing tests by adding a column argument to PreprocessNLPStage instantiation. Please let me know if I need to create a new function.

morpheus/_lib/include/morpheus/stages/preprocess_nlp.hpp
mdemoret-nv: This is more than a nitpick. Good catch. We should either be using pass-by-value and move semantics, or const-ref.

morpheus/stages/preprocess/preprocess_nlp_stage.py
mdemoret-nv: This is more what I had in mind: https://github.com/nv-morpheus/Morpheus/pull/428/files#diff-b044cf5578e03f4186927f46fd55f286a8eea240249154287735b22c989f6b43R76

morpheus/_lib/include/morpheus/stages/preprocess_nlp.hpp
```diff
@@ -54,6 +54,7 @@ class PreprocessNLPStage
                        bool truncation,
                        bool do_lower_case,
                        bool add_special_token,
+                       std::string column,
```
mdemoret-nv, line 57: Why does this come before `stride`? If it's due to the default value, we can use `"data"` for the default. If we change the order of the arguments, we need to mark this as breaking.

Can you move the `column` option to be after `stride` everywhere?

mdemoret-nv (changes requested):
Need to have the `column` argument come after `stride` otherwise this is a breaking change.

morpheus/_lib/include/morpheus/stages/preprocess_nlp.hpp
bsuryadevara: sure

bsuryadevara: @mdemoret-nv changed the order as suggested

mdemoret-nv: Great thanks.

mdemoret-nv:
rerun tests

dagardner-nv:
\rerun tests

mdemoret-nv (approved):
New tests look good. CI has been verified to complete locally without any issues.

mdemoret-nv:
@gpucibot merge

PR #436, opened 2022-11-03
Ensure Camouflage is shutdown after every test

### LINKED ISSUE #329: [BUG] camouflage not always shut down properly
**Describe the bug**
This often happens when pytest is interrupted with a Cntrl-C or a python crash. I believe another way to cause this is to launch pytest while triton is already running.

Interrupt unittests

**Expected behavior**
Shut down camouflage

mdemoret-nv:
If a test crashes or a developer stops a test in the middle of debugging, Camouflage will be left open. Any subsequent runs of pytest will fail to start.

This fixes the issue by binding the child process to the pytest process, ensuring that it will always be closed. Also added support for using already running Camouflage servers.

Closes #329

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -314,7 +314,7 @@ def wait_for_camouflage(host="localhost", port=8000, timeout=5):
 
 def _set_pdeathsig(sig=signal.SIGTERM):
     """
-    Helper function to ensure once parent process exits, its childrent processes will automatically die
+    Helper function to ensure once parent process exits, its child processes will automatically die
     """
 
     def prctl_fn():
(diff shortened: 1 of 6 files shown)


dagardner-nv (approved):
LGTM

I like the way you fixed the camouflage shutdown issue

tests/conftest.py
```diff
@@ -274,63 +277,105 @@ def chdir_tmpdir(request: pytest.FixtureRequest, tmp_path):
...
 
-    return ready
+    return False
 
 
-@pytest.fixture(scope="function")
-def launch_mock_triton():
+def _set_pdeathsig(sig=signal.SIGTERM):
+    """
+    Helper function to ensure once parent process exits, its childrent processes will automatically die
```
drobison00, line 317:
```suggestion
    Helper function to ensure once parent process exits, its child processes will automatically die
```

dagardner-nv (approved):
LGTM

mdemoret-nv:
@gpucibot merge

PR #437, opened 2022-11-03
Update Camouflage Mocks to Support gRPC

mdemoret-nv:
This PR adds mocks to Camouflage to support mocking the Triton gRPC endpoint. Currently its a pain to set up testing because we much use `MagicMock` for mocking Triton with Python and Camouflage when mocking with C++. This would unify them to both use Camouflage.

Some outstanding issues:

- [ ] This requires setting the request `id` field. Are we sure this wont have any side effects?
- [ ] What is a good way to update these payload files if they change in the future?
- [ ] What organization should we use for the payload files. Its not as clear as the HTTP folder structure
- [ ] Can we use the same payload for both HTTP and gRPC?
- [ ] The request ID solution is a bit hacky for HTTP requiring some weird tricks. Are we ok with that?


Salonijain27:
Closing this PR as it has been open for over a year. If required please reopen it.

PR #438, opened 2022-11-03
Avoid unnecessary copy in add-scores stage

dagardner-nv:
This requires ensuring that all newly built Tensors using the `create` method have their strides specified in elements not bytes.


mdemoret-nv (approved):
Looks good. Just need to merge `branch-22.11` and we should be good to go.

dagardner-nv:
@gpucibot merge

PR #457, opened 2022-11-14
Remove warning about tests not having a return value

dagardner-nv:
Removes code that is always bypassed and would fail anyways.


drobison00 (approved):
LGTM

dagardner-nv:
@gpucibot merge

PR #478, opened 2022-11-18
Improve Morpheus Shutdown Behavior On Exception

mdemoret-nv:
This PR was some outstanding work that was a result of changes in the SRF PR: https://github.com/nv-morpheus/SRF/pull/208.

If an exception is detected, we can try to manually call stop on all stages to allow for better shutdown behavior. Included in this PR are fixes to remove `CMAKE_BUILD_RPATH_USE_ORIGIN` from the CMake file which breaks building SRF with Morpheus at the same time. This option is only needed during the CI build so it was added manually to the CI scripts.


dagardner-nv (approved):
LGTM

mdemoret-nv:
@gpucibot merge

PR #535, opened 2022-12-09
Add Morpheus devcontainer

cwharris:
Adds a devcontainer capable of building Morpheus without invalidating the conda env on change.


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM ghcr.io/trxcllnt/devcontainer-images/cmake-ninja-sccache-llvm-cuda-mambaforge:latest AS base
```
drobison00, line 1: Is this something that we could put in a more official location?

.devcontainer/dev.code-workspace
```diff
@@ -0,0 +1,57 @@
...
+                    {
+                        "description": "Enable pretty-printing for gdb",
+                        "text": "-enable-pretty-printing",
+                        "ignoreFailures": true
+                    }
+                ]
+            }
+        ]
+    }
+}
```
mdemoret-nv, line 57: Can we unify the `dec.code-workspace` with the root `morpheus.code-workspace`? Having two workspace files will only be problematic.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,82 @@
...
+				"nvidia.nsight-vscode-edition",
+				"twxs.cmake",
+				"xaver.clang-format"
+			]
+        }
+    },
+    "settings": {
+        "C_Cpp.intelliSenseEngine": "disabled",
+        "python.terminal.activateEnvironment": false
+    }
```
mdemoret-nv, line 81: How do these settings interact with the settings in `.vscode/settings.json` and the settings in `morpheus.code-workspace`?

.devcontainer/bin/dev-kafka-start
```diff
@@ -0,0 +1,21 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+COMPOSE_FILE="${MORPHEUS_ROOT}/.devcontainer/docker-compose.yml"
+
+docker compose -f $COMPOSE_FILE up -d zookeeper kafka
```
mdemoret-nv, line 19: Normally, I spin up 3 instances of kafka to test multi-replica modes. Is that possible from here?

cwharris: We can just update the compose file to spin up 3 instances instead of 1. We could get fancier than that and use one compose file per service (kafka1/zookeer / kafka3/zookeeper / triton) and just pick which one to use based on a cli option, but I see no reason to do that when we can just fire up three.

.devcontainer/devcontainer.json
cwharris: From highest to lowest priority, all are considered:
```
/workspaces/morpheus/.vscode/settings.json (workspace dir) >
/workspaces/morpheus/morpheus.code-workspace (workspace file) >
/home/coder/.vscode-server/data/Machine/settings.json (devcontainer settings) >
/home/charris/.config/Code/User/settings.json (local user settings)
```

Your local user settings are considered first, then overridden by the devcontainer settings, which are overridden by the workspace file, which are then overridden by the `.vscode/settings.json` file.

So it's important that the devcontainer only contain settings that are critical to function. For instance, `python.terminal.activateEnvironment` here is important because otherwise it tries to use the wrong conda environment. `C_Cpp.intelliSenseEngine` is important because we include both the C++ and clangd plugins in the devcontainer, and clangd should take priority.

.devcontainer/dev.code-workspace
cwharris: I've deleted the devcontainer workspace, and updated the abp pcap example to save the output to a `.tmp` dir.

.devcontainer/bin/dev-kafka-start
cwharris: Is that something you want done now, or something we can update later? You may want to give some instructions on how to test your scenario.

mdemoret-nv: We can update that later if needed.

mdemoret-nv (approved):

mdemoret-nv:
@gpucibot merge

PR #555, opened 2022-12-15
Revert docs compilation instructions

cwharris:


docs/README.md
```diff
@@ -25,8 +25,9 @@ From the root of the morpheus repo:
 mamba env update -f docs/conda_docs.yml
 ```
 
-## Build Morpheus and Documentation
+## To build the documentation run:
 ```
-CMAKE_CONFIGURE_EXTRA_ARGS="-DMORPHEUS_BUILD_DOCS=ON" ./scripts/compile.sh
+cmake -B build -DMORPHEUS_BUILD_DOCS=ON .
+cmake --build build --target morpheus_docs
```
dagardner-nv, line 31: I'm curious, what you had should have worked, is this fixing a specific issue you encountered?

cwharris: i had built the documentation and mistook some warnings for errors, and thought this would fix it. I didn't notice the warnings on the second attempt, but they were indeed there, so this update does not fix anything.

PR #578, opened 2022-12-20
Update container version to 23.01 in DFP docker-compose

### LINKED ISSUE #577: [BUG]: DFP example uses older/incompat container version
### Version

23.01

Docker

### Describe the bug.

The `docker-compose.yml` builds upon `v22.09.00-runtime` which uses SRF, but the stages have been updated to use MRC.

This can be worked-around by setting `MORPHEUS_CONTAINER_VERSION`

_No response_

### Relevant log output

_No response_

### Full env printout

_No response_

### Other/Misc.

_No response_

### Code of Conduct

dagardner-nv:
Add dask

fixes #577

### DIFF
--- a/examples/digital_fingerprinting/production/conda_env.yml
+++ b/examples/digital_fingerprinting/production/conda_env.yml
@@ -23,6 +23,7 @@ channels:
 dependencies:
     ####### Morpheus Dependencies (keep sorted!) #######
     - boto3
+    - dask
     - dill
     - kfp
     - librdkafka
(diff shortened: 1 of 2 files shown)


cwharris (approved):

cwharris:
it's weird that some of the examples can use the dev container, but examples like DFP have to use a custom docker image based on a published image. it's also strange that we re-copy the example directory to the new image rather than using the example directory that is already published in the official release container.

mdemoret-nv (approved):

mdemoret-nv:
> it's weird that some of the examples can use the dev container, but examples like DFP have to use a custom docker image based on a published image. it's also strange that we re-copy the example directory to the new image rather than using the example directory that is already published in the official release container.

The DFP example is more complicated than the other examples since it requires starting other services like ML flow and we support running it from a Jupyter notebook. Because of this, it has it's own docker image and docker-compose.

dagardner-nv:
@gpucibot merge

PR #583, opened 2022-12-21
DFP: Exclude unwanted columns

### LINKED ISSUE #582: [FEA]: DFP should use a serialize stage, and exclude internal fields

Change

Low (would be nice)

The DFP output includes several fields that are internal to DFP but are not of interest to users. 

The following fields should be excluded:
* batch_count
* origin_hash
* _row_hash
* _batch_id

_No response_

### Additional context

_No response_

### Code of Conduct

dagardner-nv:
* Filter stage now optionally able to use a DataFrame column as the filter criteria.
* Choice of a DataFrame column or a Tensor is specified with a `FilterSource` enum.
* Accepted message type of the Filter stage is now driven by the choice of a DF column or Tensor as the criteria.
* `DType` now has a default constructor.
* DFP Post Processing stage no longer performs it's own filtering and no longer converts the output message to a `UserMessageMeta`.
* DFP inference pipelines updated to use Filter & Serialize stages, resulting in unwanted columns excluded from output (fixes #582).
* Update DFP docs to reflect these changes

Breaking Changes:
* `morpheus._lib.file_types` merged into `morpheus._lib.common`
* `DevMemInfo` expanded to include shape & stride attributes
* MatxUtil updated to take advantage of new attrs in `DevMemInfo`


morpheus/_lib/src/stages/filter_detection.cpp
```diff
@@ -33,126 +36,193 @@
...
+ */
+struct BufferInfo
+{
+    std::vector<std::size_t> shape;
+    std::vector<std::size_t> stride;
+    std::size_t count;
+    std::size_t bytes;
+    morpheus::DType type;
+    const uint8_t* head;
+};
```
dagardner-nv, line 61: @mdemoret-nv I wasn't sure if I should just add shape, stride & bytes to the `DevMemInfo` struct or not. 

On one hand I didn't want to make this PR any larger than it already is. However on the other hand, without these fields `DevMemInfo` many of matx_util methods which also require passing in the shape and stride.

mdemoret-nv: I think this is a good time to merge them. Keeping it separate will only lead to more work down the line. If we go down this route, we probably want to make a few changes:

1. Unify the `data` property. One uses a pointer and the other uses a `rmm::device_buffer`
2. We probably want to hold a reference to the `rmm::device_buffer` but allow a `data()` method
3. Most of these fields should be const so they dont get modified after being set. `DevMemInfo` should be immutable
4. We will probably want some utility methods to simplify getting the bytes per element and things like that.

morpheus/_lib/CMakeLists.txt
```diff
@@ -54,6 +54,9 @@ include(cmake/python_modules/common.cmake)
 #----------morpheus._lib.file_types---------
 include(cmake/python_modules/file_types.cmake)
 
+#----------morpheus._lib.filter_source---------
+include(cmake/python_modules/filter_source.cmake)
+
```
mdemoret-nv, line 59: I dont love that we have different compilation units and python modules for single enums. Can we combine this into a single module. Something like enums, data, or possible untify it with common (I forget why its separate from common)

morpheus/_lib/include/morpheus/stages/filter_detection.hpp
```diff
@@ -84,13 +84,18 @@ class FilterDetectionsStage : public mrc::pymrc::PythonNode<std::shared_ptr<Mult
      * @param threshold : Threshold to classify
      * @param copy : Whether or not to perform a copy default=true
      */
-    FilterDetectionsStage(float threshold, bool copy = true);
+    FilterDetectionsStage(float threshold,
+                          bool copy               = true,
+                          FilterSource operate_on = FilterSource::AUTO,
```
mdemoret-nv, line 89: Not a huge fan of `operate_on`. We dont have any other arguments that use an active verb like this. Can we just keep it `source`, data source, or column source or similar?

morpheus/_lib/src/python_modules/filter_source.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+namespace morpheus {
+
+namespace py = pybind11;
+
+PYBIND11_MODULE(filter_source, m)
+{
+    py::enum_<FilterSource>(
+        m, "FilterSource", "Enum to indicate which source the FilterDetectionsStage should operate on.")
+        .value("AUTO", FilterSource::AUTO)
```
mdemoret-nv, line 30: Not sure why the capitalization is different but `FileTypes` uses `Auto`, `JSON`, and `CSV`. Should we make these the same?

morpheus/_lib/src/stages/filter_detection.cpp
```diff
@@ -33,126 +36,193 @@
...
+
+BufferInfo get_tensor_buffer_info(const std::shared_ptr<morpheus::MultiMessage>& x, const std::string& field_name)
+{
+    // The pipeline build will check to ensure that our inpt is a MultiResponseMessage
+    const auto& filter_source = std::static_pointer_cast<morpheus::MultiResponseMessage>(x)->get_output(field_name);
```
mdemoret-nv, line 66: Should this just cast to a `MultiTensorMessage`?

morpheus/_lib/src/stages/filter_detection.cpp
```diff
@@ -33,126 +36,193 @@
...
+            [this, &output](sink_type_t x) {
+                BufferInfo buffer_info;
+                if (m_operate_on == FilterSource::TENSOR || m_operate_on == FilterSource::AUTO)
+                {
+                    buffer_info = get_tensor_buffer_info(x, m_field_name);
+                }
+                else
+                {
+                    buffer_info = get_column_buffer_info(x, m_field_name);
+                }
```
mdemoret-nv, line 126: `m_operate_on` shouldnt change during operation. Instead of branching inside of the hot path, return different `subscribe_fn_t` depending on the value. Then you dont need to check this value for each piece of data since it will never change.

morpheus/stages/postprocess/filter_detections_stage.py
```diff
@@ -63,18 +66,33 @@ class FilterDetectionsStage(SinglePortStage):
...
+                 operate_on: FilterSource = FilterSource.AUTO,
+                 field_name: str = "probs"):
         super().__init__(c)
 
         # Probability to consider a detection
         self._threshold = threshold
         self._copy = copy
 
+        if operate_on == FilterSource.AUTO:
+            operate_on = FilterSource.TENSOR
```
mdemoret-nv, line 91: Isnt auto supposed to look at the input data type before making this decision?

mdemoret-nv (changes requested):
Main change that we should make is improving the `DevMemInfo` class.

morpheus/_lib/CMakeLists.txt
dagardner-nv: Done, merged both of them into common, which makes this a breaking change

morpheus/_lib/include/morpheus/objects/filter_source.hpp
```diff
@@ -15,30 +15,28 @@
  * limitations under the License.
```
dagardner-nv, line 15: Not sure why git thinks this is a copy.

mdemoret-nv (approved):
Looks better using the DevMemInfo everywhere.

PR #586, opened 2022-12-30
Table locking & column preallocation

### LINKED ISSUE #456: [FEA]: Preallocate Dataframe columns needed by post-processing stages

Improvement

Medium

Some of the post-processing stages (add-class & add-scores) need to insert new columns into the Dataframe. The problem is this requires acquiring the gil and introduces a potential race condition.

Since the add-class & add-scores stages know which columns they need to create, they should be able to advertise the names & types of the columns. 

During pipeline build stage, these can be queried, and published to the source stage, so that they can be allocated in the Dataframe prior to emitting into the pipeline.

_No response_

### Additional context

Idea came out of a discussion at the Morpheus meeting as a way to avoid locking in the during pipeline execution.

### Code of Conduct

### LINKED ISSUE #490: [FEA]: DataType & DType should be merged

Improvement

Low (would be nice)

`Dtype` is a subclass of `DataType`. If I recall correctly `DataType` was originally a part of SRF, and `Dtype` is a Morpheus specific subclass of it. 

`DType`'s implementation is in a `.cu` file but the code appears to be all C++ without any cuda code.

Merge the two

_No response_

### Additional context

_No response_

### Code of Conduct

dagardner-nv:
* Builds on changes in #427
* Adds a `PreallocatorMixin` which when added to a stage performs pre-allocation. This should be added to the first stage in a pipeline which emits a DataFrame or MessageMeta in a pipeline.
* Morpheus' TypeId enum exposed to the Python API, allowing stages to define types for columns needing pre-allocation
* `MutableTableInfo` exposed to Python via a context manager to be used in `with` blocks
* `type_util` (`Dtype`) and `type_util_detail` (`DataType`) merged into a new compilation unit `dtype` fixes #490 

fixes #456


morpheus/_lib/include/morpheus/messages/meta.hpp
```diff
@@ -101,6 +99,43 @@ class MessageMeta
...
+    cudf::size_type m_stop{-1};
+    std::vector<std::string> m_column_names;
+};
+
+/****** Python Interface **************************/
+class MutableCtxMgr : public std::enable_shared_from_this<MutableCtxMgr>
+{
+  public:
+    MutableCtxMgr(MutableTableInfo&& table);
+    std::shared_ptr<MutableCtxMgr> enter();
```
mdemoret-nv, line 130: I think we can improve on this object a bit because its a bit clunky. Some ideas:

- `MutableCtxMgr` takes a `MutableTableInfo` in the constructor, meaning the table is already locked before you call `enter()`. `enter()` should be the step that does the actual locking. Instead, if the context manager accepted a `MessageMeta`, then `enter()` could actually lock the table
- `enter()` returning a pointer to itself is also weird. It kind of seems like `enter()` doesnt do much if the table is already locked and we just return a pointer to itself. Instead, we should have another object returned here
- I know there were some issues with using RAII with context managers from python, but I still think we could benefit from using that here. We should internally store an RAII object in the python wrapper. Once that object goes out of scope, it can guarantee that the table is unlocked even if the wrapper is still alive

morpheus/_lib/src/python_modules/messages.cpp
```diff
@@ -110,10 +110,17 @@ PYBIND11_MODULE(messages, m)
...
+    py::class_<MutableCtxMgr, std::shared_ptr<MutableCtxMgr>>(m, "MutableCtxMgr")
+        .def("__enter__", &MutableCtxMgr::enter, py::return_value_policy::reference)
+        .def("__exit__", &MutableCtxMgr::exit)
+        .def_property_readonly("df", &MutableCtxMgr::df_property, py::return_value_policy::reference);
+
     py::class_<MessageMeta, std::shared_ptr<MessageMeta>>(m, "MessageMeta")
         .def(py::init<>(&MessageMetaInterfaceProxy::init_python), py::arg("df"))
         .def_property_readonly("count", &MessageMetaInterfaceProxy::count)
-        .def_property_readonly("df", &MessageMetaInterfaceProxy::get_data_frame, py::return_value_policy::move)
+        .def_property_readonly("df", &MessageMetaInterfaceProxy::df_property, py::return_value_policy::move)
```
mdemoret-nv, line 121: Can we show a deprecation message when this is used? Here is an example in the MRC repo: https://github.com/nv-morpheus/MRC/pull/261/files#diff-443d85d7ff4910fdc3e77cfc71629f85257ff6f21ae0d4f23733504cbf904678R63

morpheus/_lib/src/python_modules/messages.cpp
```diff
@@ -110,10 +110,17 @@ PYBIND11_MODULE(messages, m)
     mrc::node::EdgeConnector<std::shared_ptr<morpheus::MultiResponseProbsMessage>,
                              std::shared_ptr<morpheus::MultiMessage>>::register_converter();
 
+    py::class_<MutableCtxMgr, std::shared_ptr<MutableCtxMgr>>(m, "MutableCtxMgr")
+        .def("__enter__", &MutableCtxMgr::enter, py::return_value_policy::reference)
+        .def("__exit__", &MutableCtxMgr::exit)
+        .def_property_readonly("df", &MutableCtxMgr::df_property, py::return_value_policy::reference);
```
mdemoret-nv, line 116: We should discuss, but I feel like these methods could be put on a different class or return different object. Something like the following would be nice:
```python
with meta.mutable_dataframe() as df:
    df["col"] = 4
```
Where the context manager `__enter__` returns the dataframe object directly.

mdemoret-nv (changes requested):
Completed a quick pass over the new interface. We should discuss some changes regarding the context manager.

morpheus/_lib/include/morpheus/stages/preallocate.hpp
```diff
@@ -0,0 +1,86 @@
...
+template <typename MessageT>
+class PreallocateStage : public mrc::pymrc::PythonNode<std::shared_ptr<MessageT>, std::shared_ptr<MessageT>>
+{
+  public:
+    using base_t = mrc::pymrc::PythonNode<std::shared_ptr<MessageT>, std::shared_ptr<MessageT>>;
+    using typename base_t::sink_type_t;
+    using typename base_t::source_type_t;
+    using typename base_t::subscribe_fn_t;
+
+    PreallocateStage(const std::vector<std::tuple<std::string, TypeId>>& needed_columns);
```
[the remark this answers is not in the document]
dagardner-nv: The reason why I went with a `vector` was primarily for three reasons:
1) To preserve the ordering in which they were collected. 
2) We only ever iterate over the container from start to finish, and we don't perform any look-ups.
3) The existing `insert_missing_columns` & `insert_columns` methods used a vector.

morpheus/_lib/src/python_modules/messages.cpp
dagardner-nv: We do have a deprecation warning in `meta.cpp`:
```
pybind11::object MessageMetaInterfaceProxy::df_property(MessageMeta& self)
{
    PyErr_WarnEx(
        PyExc_DeprecationWarning,
        "Warning the df property returns a copy, please use the copy_dataframe method or the mutable_dataframe "
        "context manager to modify the DataFrame in-place instead.",
        1);

    return MessageMetaInterfaceProxy::get_data_frame(self);
}
```

morpheus/_lib/include/morpheus/messages/meta.hpp
dagardner-nv: I thought about making some sort of wrapped object that looks/acts like a dataframe, but instead proxies calls to the actual dataframe something like:
```
WrappedDF():
    _df
    _ctx
    
    def __call__(self, *a, **k):
        if not self._ctx.locked:
            raise RuntimeError
        else:
            # call _df
```

The problem would be anyone who needed to call some third-party lib that expected a real proper cudf dataframe and might throw an exception on our wrapped work-alike class.

mdemoret-nv (approved):
LGTM but lets target this for 23.03

mdemoret-nv (approved):
Lets get this in early in the cycle.

Issue #554, opened 2022-12-15
[FEA]: Log parsing training for variable token lengths


raykallen:
### Is this a new feature, an improvement, or a change to existing functionality?

New Feature

### How would you describe the priority of this feature request

Critical (currently preventing usage)

### Please provide a clear description of problem this feature solves

I want Morpheus log parsing training to have desired token length for the trained model as a variable.

### Describe your ideal solution

The main functionality missing is that the token labels created from parsed logs in the training dataset need to be apply to the log pieces that are created when the length of the model is shorter than the length of the logs. 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I have searched the [open feature requests](https://github.com/nv-morpheus/Morpheus/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22%2Cimprovement%2Cenhancement) and have found no duplicates for this feature request

PR #597, opened 2023-01-05
Log parsing model training for split logs and smaller models

raykallen:
added functionality for splitting logs during training to accommodate models with shorter max sequence lengths

closes https://github.com/nv-morpheus/Morpheus/issues/554


tzemicheal (approved):
Reviewed the NB. LGTM!

PR #602, opened 2023-01-06
Remove unused `cudf_kwargs` constructor argument from FileSourceStage & NLPVizFileSource

### LINKED ISSUE #600: [BUG]: FileSourceStage cudf_kwargs is ignored
### Version

23.01

Source

### Describe the bug.

FileSourceStage saves the `cudf_kwargs` constructor argument to `self._cudf_kwargs` but doesn't actually pass it into the `read_file_to_df` function in the Python impl, and doesn't pass it to the C++ impl.

_No response_

### Relevant log output

_No response_

### Full env printout

_No response_

### Other/Misc.

_No response_

### Code of Conduct

dagardner-nv:
Arguments were being ignored. While it would have been easy to simply use these in the Python impl, it would have been difficult to implement on the C++ side.

Marking this as a breaking change as this might break API users

Fixes #600


mdemoret-nv (approved):
LGTM

PR #605, opened 2023-01-11
Update DFPMLFlowModelWriterStage to no longer save mean and std metrics

efajardo-nv:
`dfencoder` is being updated in this [PR](https://github.com/nv-morpheus/dfencoder/pull/6) to no longer store `mean` and `std` in the AutoEncoder model. This will cause an error in `DFPMLFlowModelWriterStage` where it tries to publish `mean` and `std` as metrics to MLflow. This PR updates `DFPMLFlowModelWriterStage` to no longer publish the metrics since they don't appear be used in Morpheus anymore.


hsin-c:
lgtm

mdemoret-nv (approved):

Issue #567, opened 2022-12-19
[DOC]: Document included models & datasets


dagardner-nv:
### How would you describe the priority of this documentation request

Medium

### Please provide a link or source to the relevant docs

https://github.com/nv-morpheus/Morpheus/tree/branch-23.01/models

### Describe the problems in the documentation

It looks like https://github.com/nv-morpheus/Morpheus/blob/branch-23.01/models/README.md does provide a good overview of the existing models & datasets.

We just need to:
1) Ensure its up to date
2) Include this in the sphinx build (symlink somewhere under `docs/source` to this file)
3) Update other docs, specifically the: `docs/source/getting_started.md`,  `docs/source/index.rst` and `README.md` to link to it when referring to pre-trained models. It should also appear in the table of contents on the left.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I have searched the [open documentation issues](https://github.com/nv-morpheus/Morpheus/issues?q=is%3Aopen+is%3Aissue+label%3Adoc) and have found no duplicates for this bug report

PR #612, opened 2023-01-17
Model and dataset documentation update

shawn-davis:
Fixes Issue #567


dagardner-nv:
@shawn-davis something I forgot to include in the issue (sorry for that) is that part of the request we received was to include model vram requirements for the models.

dagardner-nv (approved):

PR #627, opened 2023-01-24
Updating the CUDA libraries to use Conda packages

mdemoret-nv:
With the new devcontainer, if the default image doesnt have all of the CUDA libraries installed (cuFFT, cuRand, cuBlas, cuSolver), the build will fail. This changes the build to pull those packages in from Conda instead of assuming they are already installed.


mdemoret-nv:
This was mainly due to changing the base image for the devcontainer. Closing

Issue #618, opened 2023-01-21
[FEA]: Upgrade CI Driver Version


mdemoret-nv:
### Is this a new feature, an improvement, or a change to existing functionality?

Change

### How would you describe the priority of this feature request

High

### Please provide a clear description of problem this feature solves

The tags for the CI images have been updated. The 495 driver is being phased out in favor of the 520 version. See here: https://docs.rapids.ai/resources/github-actions/#gpu-label-combinations

### Describe your ideal solution

Upgrade our CI system to use `gpu-XXX-latest` instead of `gpu-XXX-495` to stay current with the latest CI images.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I have searched the [open feature requests](https://github.com/nv-morpheus/Morpheus/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22%2Cimprovement%2Cenhancement) and have found no duplicates for this feature request

dagardner-nv:
Related to issue #616

dagardner-nv:
@mdemoret-nv  did you resolve fix this?

mdemoret-nv:
This was fixed in PR #877

PR #634, opened 2023-01-25
Updating the CI image to use Driver 520

mdemoret-nv:
This is part 1 of 2 for updates needed by #618. For now, we are explicitly specifying 520 in the driver but plan on using `latest` in the future.


dagardner-nv (approved):
LGTM

PR #636, opened 2023-01-26
Work-around for conda-build's incorrect handling of relative submodule paths

drobison00:
Roll back to static submodule paths -- updates container release build to install pip wheel into environment instead of old style -DComponent=wheel.

<s>Issue: `conda build` doesn't handle relative submodule paths correctly, it assumes that all paths are of the form `https:`, `ssh:`, or a system path `[file_path]`. When pulling submodules it performs a naive `os.path.join(cache_dir, submodule)` call, that will reference an invalid path in the file system when submodule is a relative url path.

Complicating issues:
- Conda-build doesn't support disabling recursive submodule pulls
- If we use a simple url/file, then we don't get the GIT_XXX environment variables supplied by conda-build, which are used frequently in the meta file.
- Conda-build does a HEAD checkout, clones to a temporary directory created at run time, and attempts to pull all submodules before applying any source patches.
- We build both in and outside of a container, so modifying the source of the conda-build files is not reliable
- We can't monkey patch because conda-build is self-contained

Solution:
- Create a clone/extraction script and put it in a tarball
- Update meta.yml to use a `url` source instead of `git_url`
- Call the cloning script in morpheus_build.sh and continue build</s>


drobison00:
Closing, relevant changes were rolled into release update.

PR #659, opened 2023-01-31
Control MonitorStage output with the log-level

### LINKED ISSUE #553: [FEA]: Control MonitorStage output with the log-level

New Feature

Low (would be nice)

This would allow the `MonitorStage` to be silenced (potentially even omitted from the pipeline entirely) based on the configured log level which is typically exposed as a command line flag.

Reasoning:
The `MonitorStage` is a very useful tool for debugging & monitoring the execution of a pipeline. However it comes at a performance cost as it is a Python-only stage, and it has a tendency to clobber the output.

It would be nice to silence the monitor stages without needing to edit the pipeline script, or pipeline cli (outside of the log level flag).

Each `MonitorStage` will be optionally configured with a log-level:
```python
pipeline.add_stage(MonitorStage(config, description="AE post-proc rate", level=DEBUG)
```

And the `MonitorStage._build_single` method would be updated to do something like, which would omit the stage from the pipeline:
```python
if not logger.isEnabledFor(self._log_level):
    return input_stream
```

_No response_

### Additional context

_No response_

### Code of Conduct

dagardner-nv:
* Add new `log_level` constructor argument to `MonitorStage`
* `MonitorStage` only adds itself to the MRC pipeline if the `morpheus.stages.general.monitor_stage` logger is configured at or below `log_level`

fixes #553

### DIFF
--- a/morpheus/stages/general/monitor_stage.py
+++ b/morpheus/stages/general/monitor_stage.py
@@ -25,6 +25,7 @@ from tqdm import tqdm
 import cudf
 
 from morpheus.cli.register_stage import register_stage
+from morpheus.cli.utils import parse_log_level
 from morpheus.config import Config
 from morpheus.messages import MessageMeta
 from morpheus.messages import MultiMessage
@@ -165,7 +166,8 @@ class MonitorStage(SinglePortStage):
     determine_count_fn : typing.Callable[[typing.Any], int]
         Custom function for determining the count in a message. Gets called for each message. Allows for
         correct counting of batched and sliced messages.
-
+    log_level : str, default = "INFO"
+        Enable this stage when the configured log level is at `log_level` or lower.
     """
     stage_count: int = 0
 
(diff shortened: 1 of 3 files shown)


morpheus/stages/general/monitor_stage.py
```diff
@@ -175,7 +177,8 @@ def __init__(self,
                  smoothing: float = 0.05,
                  unit: str = "messages",
                  delayed_start: bool = False,
-                 determine_count_fn: typing.Callable[[typing.Any], int] = None):
+                 determine_count_fn: typing.Callable[[typing.Any], int] = None,
+                 log_level: str = "INFO"):
```
mdemoret-nv, line 181: `log_level` should accept an integer to work with `logging.INFO`, etc.

morpheus/stages/general/monitor_stage.py
```diff
@@ -213,12 +221,13 @@ def on_start(self):
         """
         Starts the pipeline stage's progress bar.
         """
-        # Set the monitor interval to 0 to use prevent using tqdms monitor
-        tqdm.monitor_interval = 0
+        if logger.isEnabledFor(self._log_level):
```
mdemoret-nv, line 224: Set a boolean here whether or not this stage is enabled.

morpheus/stages/general/monitor_stage.py
```diff
@@ -241,6 +250,8 @@ def _ensure_progress_bar(self):
             self._progress.reset()
 
     def _build_single(self, builder: mrc.Builder, input_stream: StreamPair) -> StreamPair:
+        if not logger.isEnabledFor(self._log_level):
```
mdemoret-nv, line 253: Check boolean here.

mdemoret-nv (approved):
Merge once comments are addressed

PR #661, opened 2023-02-01
Clean up use of find/delete for CVE

pdmack:
Minor style improvement


models/mlflow/docker/Dockerfile
```diff
@@ -33,7 +33,7 @@ RUN sed -i 's/conda activate base/conda activate mlflow/g' ~/.bashrc
 SHELL ["/opt/conda/bin/conda", "run", "-n", "mlflow", "/bin/bash", "-c"]
 
 ARG TRITON_DIR=/mlflow/triton-inference-server
-ARG TRITON_VER=r22.10
+ARG TRITON_VER=r23.01
```
mdemoret-nv, line 36: Does this require an update to all locations where the triton version is specified?

mdemoret-nv (approved):
Just one comment and then we can merge.

dagardner-nv (approved):

models/mlflow/docker/Dockerfile
mdemoret-nv: Resolving since this is only to the mlflow dockerfile.

PR #662, opened 2023-02-01
Checking out a specific commit, requires a non-shallow checkout

dagardner-nv:


mdemoret-nv (approved):
Why do we need a deep clone here? Not a big deal, just curious.

dagardner-nv:
> Why do we need a deep clone here? Not a big deal, just curious.

Without it CPM can't find the commit.

PR #673, opened 2023-02-07
[BUG] DFP benchmark related pytest error

bsuryadevara:
- Skipping DFP production directory from pytest scan
- Updated DFP benchmarks README.md with prerequisites.


dagardner-nv (approved):
LGTM

For the benchmarks themselves are they intended to be run from within the docker container? If so maybe just add the additional packages to the docker build?

PR #682, opened 2023-02-08
Move duplicated one-liner to TensorUtils

dagardner-nv:
Simple one-liner for calculating the element count of a tensor based on the shape was copy/pasted into four different places in the code base (Likely I'm to blame for at least one of these).

This just replaces them all with a call to `TensorUtils::get_elem_count`


mdemoret-nv (approved):
Looks good

PR #688, opened 2023-02-09
Fix Triton mini batch bug

### LINKED ISSUE #680: [BUG]: C++ impl for Triton inference can incorrectly split inference inputs
### Version

23.03

Docker, Conda, Source

### Describe the bug.

The Triton inference stage often needs to split up the input based on the model's max batch size, which is quite often much smaller than the the number of rows in the message (`pipeline_batch_size`), and the input is broken up into what we call a "mini-batch".

We can also have large input fields (typically variable length fields like text) which themselves are larger than the model can accept and need to be split up into multiple inference inputs, and then we perform a reduction on the multiple outputs to produce a single output for the row.

There are currently two related bugs, the first being common:
1) We perform a partition across inputs for the same row, resulting in not all outputs being reduced.
2) If a row's input is so larger that the partitioned inputs themselves are larger than the model's max batch size.

```shell
The first variation of the bug occurs with the 
`scripts/validation/abp/val-abp-all.sh` script if you remove the `--truncation=True` flag from `scripts/validation/val-run-pipeline.sh`.
```

### Relevant log output

_No response_

### Full env printout

_No response_

### Other/Misc.

_No response_

### Code of Conduct

dagardner-nv:
* Fixes issue that can occur when the number of input tensors is greater than the number of rows in the dataframe. 
* Refactor the `on_next` method of the `InferenceClientStage` which was getting a bit long and hard to follow, moving some logic to their own methods.
* Wait until all mini-batches have returned before calculating the reduction.

Fixes #680


morpheus/_lib/include/morpheus/objects/rmm_tensor.hpp
```diff
@@ -144,6 +144,13 @@ class RMMTensor : public ITensor
      */
     std::shared_ptr<ITensor> as_type(DType dtype) const override;
 
+    /**
+     * @brief Get the underlying rmm buffer as-id and does not apply the offset.
+     *
+     * @return std::shared_ptr<rmm::device_buffer>
+     */
+    std::shared_ptr<rmm::device_buffer> get_buffer();
+
```
mdemoret-nv, line 153: This makes me nervous returning a `shared_ptr` to the original rmm buffer here. I feel like there has to be a better, and less error prone way. Lets disucss.

mdemoret-nv (changes requested):

morpheus/_lib/include/morpheus/objects/rmm_tensor.hpp
dagardner-nv: Reverted

dagardner-nv:
PR #667 contained this code and was merged.

PR #691, opened 2023-02-11
Warn & replace dataframes with non-unique indexes

### LINKED ISSUE #286: [BUG] Calling set_meta with a list of strings raises an exception on a second call
**Describe the bug**
Calling `set_meta` where values is a `list` or `cudf.Series` of string values raises an exception on a second call. The first call which creates the new column succeeds while the second call, updating values over a set of sliced rows raises an exception.

Issue seems to be centered around the fact that cupy doesn't support arrays of strings, while cudf does. This bug occurs in both C++ & Python execution modes.

Once this is fixed we need to remove the work-around in `examples/abp_pcap_detection/abp_pcap_preprocessing.py` on line 176.

Causes #282

```python
import os
import string

from morpheus._lib.file_types import FileTypes
from morpheus.io.deserializers import read_file_to_df
from morpheus.messages.multi_message import MultiMessage
from morpheus.messages.message_meta import MessageMeta

input_file = os.path.join(os.getenv('MORPHEUS_ROOT', '.'), 'tests/tests_data/filter_probs.csv')

df = read_file_to_df(input_file, file_type=FileTypes.Auto, df_type='cudf')
meta = MessageMeta(df)
mm1 = MultiMessage(meta, 0, 5)
mm2 = MultiMessage(meta, 5, 5)

values = list(string.ascii_letters)

mm1.set_meta('letters', values[0:5])
mm2.set_meta('letters', values[5:10])
```

### LINKED ISSUE #393: [BUG] Starter DFP error with increased inference input size
**Describe the bug**
The following error is seen with the Starter CloudTrail DFP example pipeline when inference input is increased 100x.
```
AssertionError: Gaps in blk ref_locs

At:
  /opt/conda/envs/morpheus/lib/python3.8/site-packages/pandas/core/internals/managers.py(240): _rebuild_blknos_and_blklocs
  /opt/conda/envs/morpheus/lib/python3.8/site-packages/pandas/core/internals/managers.py(627): _consolidate_inplace
  /opt/conda/envs/morpheus/lib/python3.8/site-packages/pandas/core/internals/managers.py(666): reindex_indexer
  /opt/conda/envs/morpheus/lib/python3.8/site-packages/pandas/core/generic.py(4883): _reindex_with_indexers
  /opt/conda/envs/morpheus/lib/python3.8/site-packages/pandas/core/indexing.py(1063): _multi_take
  /opt/conda/envs/morpheus/lib/python3.8/site-packages/pandas/core/indexing.py(1107): _getitem_tuple
  /opt/conda/envs/morpheus/lib/python3.8/site-packages/pandas/core/indexing.py(925): __getitem__
  /my_data/gitrepos/efajardo-nv/Morpheus/morpheus/messages/multi_message.py(115): get_meta
  /my_data/gitrepos/efajardo-nv/Morpheus/morpheus/stages/preprocess/preprocess_ae_stage.py(88): pre_process_batch
```

Run CloudTrail DFP Pipeline CLI example from [README](https://github.com/nv-morpheus/Morpheus/blob/branch-22.09/examples/digital_fingerprinting/starter/README.md) with `--repeat=100` option:
```
morpheus --log_level=DEBUG \
run --num_threads=1 --pipeline_batch_size=1024 --model_max_batch_size=1024 --use_cpp=False \
pipeline-ae \
--columns_file=morpheus/data/columns_ae_cloudtrail.txt \
--userid_column_name=userIdentitysessionContextsessionIssueruserName \
--userid_filter=user123 \
--feature_scaler=standard \
from-cloudtrail \
--input_glob=models/datasets/validation-data/dfp-cloudtrail-*-input.csv \
--max_files=200 \
--repeat=100 \
train-ae \
--train_data_glob=models/datasets/training-data/dfp-cloudtrail-*.csv \
--source_stage_class=morpheus.stages.input.cloud_trail_source_stage.CloudTrailSourceStage \
--seed=42 \
preprocess \
inf-pytorch \
add-scores \
serialize \
to-file --filename=./cloudtrail-dfp-detections.csv --overwrite
```

**Expected behavior**
Pipeline to complete successfully with no errors.

**Environment overview (please complete the following information)**
 - Environment location: Docker
 - Method of Morpheus install: Docker
   - Docker DEV container: morpheus:dev-220929

### LINKED ISSUE #626: [BUG]: `AppShieldSourceStage.read_file_to_df` shows an error message even if the function succeeds
### Version

23.01a

Source

### Describe the bug.

When running the test: `./tests/test_appshield_source_stage.py::test_read_file_to_df[expected_df0-cols_exclude0]`, you will see output like the following:

```
tests/test_appshield_source_stage.py Error while loading file content to datframe with 'cols_exclude' filter
Traceback (most recent call last):
  File "/opt/conda/envs/morpheus/lib/python3.8/site-packages/pandas/core/internals/construction.py", line 906, in _finalize_columns_and_data
    columns = _validate_or_indexify_columns(contents, columns)
  File "/opt/conda/envs/morpheus/lib/python3.8/site-packages/pandas/core/internals/construction.py", line 954, in _validate_or_indexify_columns
    raise AssertionError(
AssertionError: 2 columns passed, passed data had 5 columns

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/workspaces/morpheus-dev2/morpheus/stages/input/appshield_source_stage.py", line 166, in read_file_to_df
    plugin_df = pd.DataFrame(columns=features_plugin, data=data["data"])
  File "/opt/conda/envs/morpheus/lib/python3.8/site-packages/pandas/core/frame.py", line 694, in __init__
    arrays, columns, index = nested_data_to_arrays(
  File "/opt/conda/envs/morpheus/lib/python3.8/site-packages/pandas/core/internals/construction.py", line 483, in nested_data_to_arrays
    arrays, columns = to_arrays(data, columns, dtype=dtype)
  File "/opt/conda/envs/morpheus/lib/python3.8/site-packages/pandas/core/internals/construction.py", line 807, in to_arrays
    content, columns = _finalize_columns_and_data(arr, columns, dtype)
  File "/opt/conda/envs/morpheus/lib/python3.8/site-packages/pandas/core/internals/construction.py", line 909, in _finalize_columns_and_data
    raise ValueError(err) from err
ValueError: 2 columns passed, passed data had 5 columns
```

Which is misleading because the function actually has a try/except block and handles the error correctly. We should make the following changes:

1. Print a warning instead of an error here or remove it altogether
2. Try to handle the data better to avoid the error in the first place. This would be handled by checking the type of `data["data"]`. If it is a dictionary, there is nothing to do. If its a list, then we need to look up the positions of the columns in `titles` and filter the indexes from the array. This way, you dont need to load the entire dataframe and then exclude columns.

```shell
pytest -s -v ./tests/test_appshield_source_stage.py::test_read_file_to_df\[expected_df0-cols_exclude0\]
```
```


### Relevant log output

_No response_

### Full env printout

_No response_

### Other/Misc.

_No response_

### Code of Conduct

### LINKED ISSUE #686: [FEA]: Detect non-unique DataFrame index objects and generate an error/warning

Improvement

High

If you run a pipeline using a DataFrame that has a non-unique index, it breaks the `MutliMessage` slicing operations in subtle ways which can cause unintended bugs. The root cause of the issue is due to this code in `MultiMessage.get_meta()`:

```python
idx = self.meta.df.index[self.mess_offset:self.mess_offset + self.mess_count]
return self.meta.df.loc[idx, :]
```
If `self.meta.df.index` is non-unique, then each element in `idx` can reference more than one row resulting in `self.meta.df.loc[idx, :]` containing more rows than `self.mess_count`.

When this happens, the pipeline will continue on as normal introducing indexing errors down the line.

Ideally, we should detect this and either fix it or error out when it happens. The best places to handle this would be:

1. In the `DeserializeStage` body, we could easily detect if `x.meta.df.index.is_unique == True`
   1. At this point in the pipeline, we should just generate a warning and then fix up the index
   2. This is a good point to fix the DF since it hasnt been converted to `MultiMessage` yet
   3. This fix would cover most pipelines but not all since not all use `DeserializeStage`
2. In `MultiMessage.get_meta()` we could check to make sure `self.meta.df.index.is_unique == True`
   1. At this point, we would just want to error since we do not have exclusive access to the DF
   2. Performing the check here would guarantee that we detect the issue before it is about to occur

The following code fixes the index while preserving the original values in a column named `_index_`. This pattern has been used in DFP and for some validation pipelines, but should be standardized across the board (using utility functions and standard column names).

```python
# Reset the index preserving the original index in a new column
df.index.name = "_index_" + (df.index.name or "")
df.reset_index(inplace=True)
```

_No response_

### Additional context

_No response_

### Code of Conduct

### LINKED ISSUE #687: [FEA]: Improve developer experience with `MessageMeta` and `MultiMessage` classes

Improvement

High

How to correctly use the `MessageMeta`, `MultiMessage` and `TensorMemory` classes is a common error I see from developers when making custom pipelines. The main issue seems to stem around which values to use for `offset`, `count`, `mess_offset` and `mess_count` when changing or creating new `MultiMessage` classes.

Related to #674

We could improve the developer experience in a few ways:

1. Better documentation in the developer guide
   1. We have few common scenarios that would be good to point out when creating a new `MultiMessage` object:
      1. If using an existing `MessageMeta`:
         1. In this scenario, you would want to copy the `mess_offset` and `mess_count` from the incoming message's `MessageMeta`.
      1. If creating a new `MessageMeta` at the same time:
         1. In this scenario, `mess_offset = 0` and `mess_count = MessageMeta.count` 
      1. If using an existing `TensorMemory`:
         1. In this scenario, you would want to copy the `offset` and `count` from the incoming message's `TensorMemory`.
      1. If creating a new `TensorMemory` at the same time:
         1. In this scenario, `offset = 0` and `count = TensorMemory.count`
9. Adding static functions with more overloads to create these classes for the developer in different scenarios
   1. One overload could be made for each of the above scenarios to remove the need for developers to manually set the values
10. Adding assertions to the `MessageMeta` and `MultiMessage` classes to check for common mistakes
    1. For example, one common mistake is choosing `offest` and `count` options which are impossible. If the `MessageMeta` class only has 100 rows, then the following must be true `MultiMessage.offset + MultiMessage.count <= MessageMeta.count`. Otherwise you can index past the last item in the meta.

_No response_

### Additional context

_No response_

### Code of Conduct

### LINKED ISSUE #689: [BUG]: Setting `include_index_col=False` on WriteToFileStage fails assert at table_info.cpp:240
### Version

23.03

Source

### Describe the bug.

Happens when the output format is json. Not sure if this is just my current branch or happens in branch-23.03 as well.

_No response_

### Relevant log output

_No response_

### Full env printout

_No response_

### Other/Misc.

_No response_

### Code of Conduct

dagardner-nv:
* Add `has_unique_index` & `replace_non_unique_index` helper methods to `MessageMeta`
* Create a bunch of unittests
* Only use the index for slicing if it is unique, otherwise use a boolean mask
* The `file_type` argument of `read_file_to_df` now has a default value of `Auto`
* `DeserializeStage` checks for non-unique indexes and replaces them if needed. 
 
This comes at a performance cost in that the `DeserializeStage` needs to acquire the GIL in order to check if the Dataframe has a unique index, impacting users who never run into this issue. We could work around this by providing a `no-check` argument to the stage, or we could do the check in the constructor of `MessageMeta` when we already have the GIL and can perform the check quite cheapely.

Fixes #689
Fixes #686
Fixes #687 
Fixes #286
Fixes #626 
Fixes #393


mdemoret-nv (changes requested):
Change the destructor of `MutableTableInfo` to `LOG(ERROR)` instead of `LOG(FATAL)`.

morpheus/_lib/include/morpheus/messages/meta.hpp
```diff
@@ -68,6 +69,23 @@ class MessageMeta
...
+    bool has_sliceable_index() const;
+
+    /**
+     * @brief Replaces the index in the underlying dataframe if the existing one is not unique and monotonic. The old
+     * index will be preserved in a column named `_index_{old_index.name}`. If `has_sliceable_index() == true`, this is
+     * a no-op.
+     *
+     * @return std::string The name of the column with the old index or nullopt if no changes were made.
+     */
+    virtual std::optional<std::string> ensure_sliceable_index();
```
dagardner-nv, line 87: I'm assuming we're returning optional since the index often doesn't have a name and an empty string often is the name of the index?

morpheus/_lib/include/morpheus/utilities/cudf_util.hpp
```diff
@@ -32,30 +31,43 @@ namespace morpheus {
...
+     */
+    static pybind11::object table_from_table_with_metadata(cudf::io::table_with_metadata&& table, int index_col_count);
+
+    /**
+     * @brief Converts a C++ TableInfo (which is a view into a python table) into a Python DataTable object
+     *
+     * @param table_info C++ table view
+     * @return pybind11::object
+     */
+    static pybind11::object table_from_table_info(const morpheus::TableInfoBase& table_info);
```
dagardner-nv, line 59: So does this copy the underlying data, or just a new python object wrapping the libcudf table?

morpheus/_lib/src/messages/multi.cpp
```diff
@@ -284,13 +362,43 @@ void MultiMessageInterfaceProxy::set_meta(MultiMessage& self, pybind11::object c
...
+    auto [row_indexer, column_indexer] = get_indexers(self, df, columns);
+
+    // Check to see if this is adding a column. If so, we need to use .loc instead of .iloc
+    if (column_indexer.contains(-1))
+    {
+        // cudf is really bad at adding new columns. Need to use loc with a unique and monotonic index
+        py::object saved_index = py::none();
+
+        // Check to see if we can use slices
+        if (!df.attr("index").attr("is_unique").cast<bool>() || !df.attr("index").attr("is_monotonic").cast<bool>())
```
dagardner-nv, line 374: I'm wondering if there would be any performance benefit to saving `df.attr("index")` to a variable like:
```
auto idx_attr = df.attr("index");
if (!idx_attr.attr("is_unique").cast<bool>() || !idx_attr.attr("is_monotonic").cast<bool>())
{
    saved_index = std::move(idx_attr);
```

morpheus/_lib/src/messages/multi_tensor.cpp
```diff
@@ -80,18 +105,37 @@ void MultiTensorMessage::get_slice_impl(std::shared_ptr<MultiMessage> new_messag
...
+    if (start < 0 || start >= this->count)
+    {
+        throw std::out_of_range("Invalid memory `start` argument");
+    }
+
+    // Stop must be between (start, mess_count]
+    if (stop <= start or stop > this->count)
+    {
+        throw std::out_of_range("Invalid memory `stop` argument");
+    }
```
dagardner-nv, line 118: If I understand correctly this means start & stop are always absolute, and not relative to the offset?

morpheus/_lib/src/objects/table_info.cpp
```diff
@@ -97,7 +102,7 @@ TableInfoData TableInfoData::get_slice(cudf::size_type start,
 
                            if (found_col == this->column_names.end())
                            {
-                               throw std::runtime_error("Unknown column: " + c);
+                               throw py::key_error("Unknown column: " + c);
```
dagardner-nv, line 105: Shouldn't this be caught in the python wrappers and re-thrown as a key_error, or is this function only ever called by Python?

morpheus/_lib/src/messages/multi_tensor.cpp
```diff
@@ -80,18 +105,37 @@ void MultiTensorMessage::get_slice_impl(std::shared_ptr<MultiMessage> new_messag
...
 
-    sliced_message->offset = start;
+    // Start must be between [0, mess_count)
+    if (start < 0 || start >= this->count)
+    {
+        throw std::out_of_range("Invalid memory `start` argument");
+    }
+
+    // Stop must be between (start, mess_count]
+    if (stop <= start or stop > this->count)
```
dagardner-nv, line 115:
```suggestion
    if (stop <= start || stop > this->count)
```

morpheus/_lib/include/morpheus/objects/tensor_object.hpp
```diff
@@ -163,6 +163,8 @@ struct ITensor : public ITensorStorage, public ITensorOperations
 
     virtual TensorIndex stride(TensorIndex) const = 0;
 
+    virtual intptr_t stream() const = 0;
```
dagardner-nv, line 166: I like this, but it is a slightly different direction than I took in https://github.com/nv-morpheus/Morpheus/pull/730

Where I filled in the `MemDescriptor` struct to hold the stream pointer.

morpheus/_lib/include/morpheus/messages/meta.hpp
mdemoret-nv: Not exactly, `ensure_sliceable_index` is something you can just call and its a no-op if there is nothing to do (this means you dont have to call `has_sliceable_index` before calling `ensure_sliceable_index` since we do that anyways). However, if you want to know if something did happen, thats what the optional is for. In the docstring `The name of the column with the old index or nullopt if no changes were made`

morpheus/_lib/include/morpheus/utilities/cudf_util.hpp
mdemoret-nv: Its a python object wrapping the table. We were doing something similar in `copy_to_py_object` but that was using `loc` which is problematic. This gets rid of that and centralizes the conversion code in one place

morpheus/_lib/src/messages/multi.cpp
mdemoret-nv: Fixed

morpheus/_lib/src/messages/multi_tensor.cpp
mdemoret-nv: No start and stop are relative to the offset. But you are still bound by the relative count. Before you could call `get_slice(10, 20).get_slice(0, 20)` which is inconsistent because you are trying to get 20 rows from a 10 row slice. This just enforces that slices must be equal to or smaller than the parent.

morpheus/_lib/src/objects/table_info.cpp
mdemoret-nv: `py::key_error` is still just a C++ error. It derives from `std::runtime_error`. The benefit here is that if its caught by pybind11, then it can be translated into a Python error. Otherwise its just a normal C++ exception.

morpheus/_lib/src/messages/multi_tensor.cpp
mdemoret-nv: Fixed

morpheus/_lib/include/morpheus/objects/tensor_object.hpp
mdemoret-nv: This was a quick fix to add the stream in the least amount of changes. We can look at a better implementation in #730

mdemoret-nv:
After many additions to this PR, here is the final list of changes.

## Breaking Changes

- Changes to `MultiMessage`
  - Consistency checks are now performed on creation. This will raise errors for invalid offset and count configurations. Such as:
    - `meta` cannot be `None`
    - `mess_offset` must be in the range `[0, meta.count)`
    - `mess_count` must be in the range `(0, meta.count - mess_offset]`
  - Derived classes must define their own `__init__` to enforce keyword-only arguments
    - More info [here](https://medium.com/@aniscampos/python-dataclass-inheritance-finally-686eaf60fbb5)
  - `get_meta` and `set_meta` use `.iloc` instead of `.loc`
    - `.iloc` is faster but less flexible. This may cause some issues with uncommon data types
  - Bounds are enforced on `get_slice`
    - Before, there was no checking on the `start`/`stop` parameters. So you could get a slice that was larger than your current one
    - Now, slicing ensures that `start` is `[0, mess_count)` and `stop` is `(start, mess_count]`. Errors are thrown otherwise
  - All `__init__` parameters must be passed by keyword
    - This allows properly creating the right derived types in `get_slice` and `copy_ranges`. Before each derived type needed to implement these individually, leading to multiple conflicting implementations
- Changes to `MultiTensorMessage`
  - Consistency checks are now performed on creation. This will raise errors for invalid offset and count configurations. Such as:
    - `memory` cannot be `None`
    - `offset` must be in the range `[0, memory.count)`
    - `count` must be in the range `(0, memory.count - offset]`
    - `count` must be >= `mess_count`
  - If a `seq_ids` tensor is supplied on creation, it must have absolute message IDs instead of relative. This is to standardize whether `seq_ids` is relative to `mess_offset` or not. This also guarantees the above checks will not fail due to incorrect `seq_ids`. This check amounts to:
    - The first element, `seq_ids[offset]` must be equal to `mess_offset`
    - The last element, `seq_ids[offset + count - 1]` must be equal to `mess_offset + mess_count`
  - Bounds are enforced on `get_slice`
    - Similar to the bounds check on `MultiMessage`, this checks `count` and `offset`

### Additional Changes

- C++ and Python message classes have significantly improved tests to ensure consistent behavior
  - This was responsible for fixing several issues listed above
  - Tests are also parameterized on cudf/pandas and C++/Python to increase coverage
  - Tests now check non-unique and non-monotonic index types
- Corrent inheritance has been implemented for all `MultiInferenceXXX` and `MultiResponseXXX` messages
- New `Tensor.from_cupy()` and `Tensor.to_cupy()` to test converting between Python/C++ data structures
- New `MultiMessage.from_message()` utility class to streamline creating messages from other messages
  - This allows for things like `MultiInferenceMessage.from_message(incoming_message, count=10)` where all properties from `incoming_message` will be used to create a new `MultiInferenceMessage`. Additional supplied keyword arguments, `count` in the example, will override the values in the from message, `incoming_message`
  - This should be used as much as possible to set consistent offset and count values
- All `MultiMessage` implementations now take default values for `mess_offset`, `mess_count`, `offset`, `count` to make creation easier

mdemoret-nv (approved):

Issue #700, opened 2023-02-13
[BUG]: KafkaSourceStage doesn't stop reading until the queue is empty


dagardner-nv:
### Version

23.03

### Which installation method(s) does this occur on?

Docker, Conda, Source

### Describe the bug.

For a large queue this means that the system will always run out of memory before it emits a meessage.
It really should emit a MessageMeta every time it reaches the `pipeline_batch_size`.

### Code of Conduct

- [X] I agree to follow Morpheus' Code of Conduct
- [X] I have searched the [open bugs](https://github.com/nv-morpheus/Morpheus/issues?q=is%3Aopen+is%3Aissue+label%3Abug) and have found no duplicates for this bug report

dagardner-nv:
Bug doesn't exist. The Kafka source already chunks according to the `pipeline_batch_size`

PR #710, opened 2023-02-14
Add unittest to ensure the kafka source is respecting the pipeline batch size

dagardner-nv:
I thought this was a bug #700 so I wrote a test, and confirmed that this wasn't actually a bug, but don't want to waste a good unittest.


mdemoret-nv (approved):
Love a good test.

PR #712, opened 2023-02-15
[FEA] Create Broadcast Stage

bsuryadevara:
- Implemented BroadcastStage and incorporate to DFP production example.
- Added tests for BroadcastStage.


examples/digital_fingerprinting/production/morpheus/dfp/utils/derive_args.py
```diff
@@ -161,14 +176,31 @@ def _set_mlflow_tracking_uri(self):
         mlflow.set_tracking_uri(self._tracking_uri)
         logger.info("Tracking URI: %s", mlflow.get_tracking_uri())
 
+    def _set_time_fields(self):
+        if self._is_train_and_infer or self._is_training or self._is_inference:
+            self._time_fields = self._create_time_fields(self._duration)
+        else:
+            raise Exception(
```
dagardner-nv, line 183: In general I think raising the base `Exception` class should be avoided.

```suggestion
            raise ValueError(
```

examples/digital_fingerprinting/production/morpheus/dfp_modules_pipeline.py
```diff
@@ -0,0 +1,187 @@
...
+
+    if len(dfp_deployment_stage.output_ports) > 1:
+        tra_mntr_stage = MonitorStage(config, description="DFPTraining Pipeline rate", smoothing=0.001)
+        inf_mntr_stage = MonitorStage(config, description="DFPInference Pipeline rate", smoothing=0.001)
+
+        tra_mntr_stage = pipeline.add_stage(tra_mntr_stage)
+        inf_mntr_stage = pipeline.add_stage(inf_mntr_stage)
+
+        pipeline.add_edge(dfp_deployment_stage.output_ports[0], tra_mntr_stage)
+        pipeline.add_edge(dfp_deployment_stage.output_ports[1], inf_mntr_stage)
```
dagardner-nv, line 176: Maybe I'm just not familiar enough with modules yet, but how do you know that when the module has 2 ports, port 0 will be training and port 1 will be inference?

I wonder if there a way to attach some properties to the module, or if there is a way to look this up in the config?

examples/digital_fingerprinting/production/morpheus/dfp/modules/dfp_deployment.py
```diff
@@ -0,0 +1,87 @@
...
+    module_config = get_module_config(DFP_DEPLOYMENT, builder)
+
+    preproc_conf = module_config.get(DFP_PREPROC, None)
+    infer_conf = module_config.get(DFP_INF, None)
+    train_conf = module_config.get(DFP_TRA, None)
+
+    if "output_port_count" not in module_config:
+        raise Exception("Missing required attribute 'output_port_count'")
+
+    output_port_count = module_config.get("output_port_count")
```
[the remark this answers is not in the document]
mdemoret-nv: @dagardner-nv `AttributeError` is a very specific error which means you tried to access a property that does not exist. `KeyError` here is appropriate. I would just chain the exception like this:

```python
    try:
        output_port_count = module_config["output_port_count"]
    except KeyError as ex:
        raise KeyError("Missing required attribute 'output_port_count' on module config") from ex
```

tests/test_broadcast_stage.py
```diff
@@ -0,0 +1,52 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from unittest import mock
+
+import pytest
+
+from morpheus.stages.general.broadcast_stage import BroadcastStage
```
mdemoret-nv, line 21: `test_broadcast_stage` should have a test which uses the node in a pipeline and verifies the values it outputs are correct

morpheus/stages/general/broadcast_stage.py
```diff
@@ -0,0 +1,97 @@
...
+from mrc.core.node import Broadcast
+
+from morpheus.config import Config
+from morpheus.pipeline.stage import Stage
+from morpheus.pipeline.stream_pair import StreamPair
+
+logger = logging.getLogger(__name__)
+
+
+class BroadcastStage(Stage):
```
mdemoret-nv, line 28: So a `BroadcastStage` should only be a single port stage. It's output port is one that can be used to connect multiple downstream stages too. But they all connect to the same port. You would use it like this:

```python
pipe = Pipeline(config)
source_stage = pipe.add_stage(FileSourceStage(config, filename=input_file, iterative=False))

broadcast_stage = pipe.add_stage(BroadcastStage(config))

write_to_file_stage1 = pipe.add_stage(WriteToFileStage(config, filename=out_file1, overwrite=False))
write_to_file_stage2 = pipe.add_stage(WriteToFileStage(config, filename=out_file2, overwrite=False))

pipe.add_edge(source_stage, broadcast_stage)

pipe.add_edge(broadcast_stage, write_to_file_stage1)
pipe.add_edge(broadcast_stage, write_to_file_stage2)

pipe.run()
```
So this stage can inherit from `SinglePortStage` and doesnt need to use ports at all.

morpheus/stages/general/broadcast_stage.py
```diff
@@ -0,0 +1,97 @@
...
+        typing.Tuple
+            Accepted input types.
+
+        """
+        return (typing.Any, )
+
+    def _get_broadcast_node(self, builder) -> Broadcast:
+        # Create a broadcast node
+        node = Broadcast(builder, "broadcast")
+        return node
```
mdemoret-nv, line 77: This is a C++ node. So `supports_cpp_node` should return True

morpheus/stages/general/nonlinear_modules_stage.py
```diff
@@ -0,0 +1,119 @@
...
+
+        builder.make_edge(in_stream_node, mod_in_stream)
+
+        out_stream_pairs = []
+
+        count = 0
+        while (count < self._output_port_count):
+            out_port = f"{self._output_port_name_prefix}-{count}"
+            out_stream_pairs.append((module.output_port(out_port), self._ouput_type))
+            count += 1
```
mdemoret-nv, line 117: This seems cumbersome and enforces a lot of rules on the port naming scheme. A better implementation would be to accept 2 lists in the constructor: `input_port_names` and `output_port_names`. Then in the constructor you can do:

```python
self._create_ports(len(input_port_names), len(output_port_names))
```
Then in the build function, assign the correct `in_stream_pairs` to the correct input ports using `input_port_names`. And then return the `out_stream_pairs` just like you are doing now.

This would support any port naming scheme and would be very flexible.

morpheus/stages/general/nonlinear_modules_stage.py
```diff
@@ -0,0 +1,119 @@
...
+
+from morpheus.config import Config
+from morpheus.pipeline.stage import Stage
+from morpheus.pipeline.stream_pair import StreamPair
+from morpheus.utils.module_utils import load_module
+
+logger = logging.getLogger(__name__)
+
+
+class NonLinearModulesStage(Stage):
```
mdemoret-nv, line 28: This name isnt great. This class really wraps a multi-port MRC module in a stage and hooks up the ports. Somethink like `MultiPortModuleStage` would be better.

morpheus/stages/general/nonlinear_modules_stage.py
```diff
@@ -0,0 +1,119 @@
...
+    input_port_name : str
+        Name of the input port for the registered module.
+    output_port_name_prefix : str
+        Prefix name of the output ports for the registered module.
+    output_port_count : str
+        Number of output ports for the registered module.
+    input_type : default `typing.Any`
+        The stage acceptable input type.
+    output_type : default `typing.Any`
+        The output type that the stage produces.
```
mdemoret-nv, line 47: All of these parameters should be reduced down to `input_port_pairs` and `output_port_pairs` which are `typing.List[typing.Tuple[str, typing.Type]]`. 

The current design limits the number of input ports to 1 and limits the output types to all be the same when this really isnt a needed restriction.

mdemoret-nv (changes requested):
Couple of things that need to be cleaned up:

1. `BroadcastStage` needs to be changed to a single port stage
2. Tests need to be added to verify the values out of `BroadcastStage`
3. Tests need to be added for `non_linear_modules_stage` (Or whatever we end up calling it)
4. Does the DFP pipeline use the `BroadcastStage` anywhere?

PR #714, opened 2023-02-16
Fix detecting index column when reading from CSV in C++

### LINKED ISSUE #690: [BUG]: Deserializing a dataframe from C++ never detects an index
### Version

23.03

Source

### Describe the bug.

This check:
```
    if (data_table.metadata.schema_info.size() >= 1 &&
        data_table.tbl->get_column(0).type().id() == cudf::type_id::INT64)
```
Always fails. 
For CSV data `schema_info.size()` is always 0. 
For JSON data the `schema_info.size()` will be the number of columns, but `columns` will be `1` presumably representing the object so the type is never `INT64`.

_No response_

### Relevant log output

_No response_

### Full env printout

_No response_

### Other/Misc.

_No response_

### Code of Conduct

dagardner-nv:
* Fix logic in `get_index_col_count` in determining if the incoming DF has an index
* C++ tests for deserializers
* Expose the C++ deserializers to python

fixes #690


morpheus/_lib/tests/test_deserializers.cpp
```diff
@@ -0,0 +1,55 @@
...
+            EXPECT_EQ(get_index_col_count(table), 0);
+        }
+    }
+
+    {
+        // now test a file with an index
+        auto input_file = morpheus_root / "tests/tests_data/filter_probs_w_id_col.csv";
+        auto table      = load_table_from_file(input_file);
+        EXPECT_EQ(get_index_col_count(table), 1);
+    }
```
mdemoret-nv, line 54: Make these separate tests

morpheus/_lib/tests/test_deserializers.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include <filesystem>
+#include <string>
+#include <vector>
+
+using namespace morpheus;
+
+TEST_CLASS(Deserializers);
+
+TEST_F(TestDeserializers, GetIndexColCount)
```
mdemoret-nv, line 33: We should also add tests here that verify round trip consistency. That we can read then write a file exactly as it was originally.

morpheus/_lib/src/io/deserializers.cpp
```diff
@@ -111,23 +111,23 @@ cudf::io::table_with_metadata load_table_from_file(const std::string& filename)
 int get_index_col_count(cudf::io::table_with_metadata& data_table)
 {
     int index_col_count = 0;
+    auto& col_names     = data_table.metadata.column_names;
 
     // Check if we have a first column with INT64 data type
-    if (data_table.metadata.schema_info.size() >= 1 &&
-        data_table.tbl->get_column(0).type().id() == cudf::type_id::INT64)
+    if (col_names.size() >= 1 && data_table.tbl->get_column(0).type().id() == cudf::type_id::INT64)
```
mdemoret-nv, line 117: We need a test that has integer columns in the input dataset to make sure we arent over zealous here

morpheus/_lib/src/io/deserializers.cpp
```diff
@@ -111,23 +111,23 @@ cudf::io::table_with_metadata load_table_from_file(const std::string& filename)
 int get_index_col_count(cudf::io::table_with_metadata& data_table)
```
mdemoret-nv, line 111: Dont love that this mutates the input argument. We should make a second method called `prepare_df_index` or something similar that does the index rename and also returns the index count. Then you can call `int get_index_col_count(const cudf::io::table_with_metadata& data_table)` if you just need the index count and `int prepare_df_index(cudf::io::table_with_metadata& data_table)` if you need to get the count and update the column name.

morpheus/messages/message_meta.py
```diff
@@ -80,6 +82,13 @@ def __init__(self, df: pd.DataFrame) -> None:
         self._mutex = threading.RLock()
         self._df = df
 
+    @classmethod
+    def make_from_file(cls, filename: str):
+        if (cls._should_use_cpp()):
+            return cls._cpp_class.make_from_file(filename)
+        else:
+            return MessageMeta(read_file_to_df(filename, file_type=FileTypes.Auto, df_type="cudf"))
```
mdemoret-nv, line 90: Why is this necessary? Shouldnt creating `MessageMeta` pick C++/Python on its own?

mdemoret-nv, line 90: This determines too much for the user. They cant change the options to `read_file_to_df`. Whats the purpose of this function again? Is it used anywhere?

mdemoret-nv, line 90: A class method should call `cls(read_file_to_df(filename, file_type=FileTypes.Auto, df_type="cudf"))` to return the class we are calling this on. i.e. `UserMessageMeta.make_from_file()` would return a `MessageMeta` with the way it is.

morpheus/messages/message_meta.py
```diff
@@ -80,6 +82,13 @@ def __init__(self, df: pd.DataFrame) -> None:
         self._mutex = threading.RLock()
         self._df = df
 
+    @classmethod
+    def make_from_file(cls, filename: str):
```
mdemoret-nv, line 86: Just not sure I love this method. It doesnt allow you to create any other classes besides `MessageMeta`. Also it doesnt forward on arguments to the constructor of the class type. And finally, it would prevent using the argument `filename` in derived constructors if you worked around the other issues.

mdemoret-nv (changes requested):
Couple of things to discuss about these changes

morpheus/messages/message_meta.py
dagardner-nv: Because it's a `classmethod`, and needs to determine the impl the same way the `__new__` method does.

dagardner-nv: The `MessageMetaInterfaceProxy` class has had this method:
```cpp
static std::shared_ptr<MessageMeta> init_cpp(const std::string& filename);
```

For some time, but it was never mapped to anything in the python bindings, so I figure I might as well go ahead and add the binding.

It has the advantage of being able to, from Python, deserialize a dataframe from file and construct a `MessageMeta` entirely in C++.

Without this, the user would have to do the file read in Python (`read_file_to_df`).

morpheus/messages/message_meta.py
dagardner-nv: Yeah, that's fair. The alternative would be to expose the C++ deserializers to Python, such that `read_file_to_df` will call the C++ impl when `df_type="cudf"`.

The down side is it would be an extra trip back to python to return the dataframe, and then back in to C++ to construct the `MessageMeta`.

Personally I do a lot of testing/debugging from `ipython` and I typically need to do:
```python
from morpheus._lib.common import FileTypes
from morpheus.io.deserializers import read_file_to_df
from morpheus.messages import MessageMeta

m = MessageMeta(read_file_to_df('tests/tests_data/filter_probs.csv', file_type=FileTypes.Auto, df_type='cudf'))
```

With this `classmethod` that trims down to:
```python
from morpheus.messages import MessageMeta
m = MessageMeta.make_from_file('tests/tests_data/filter_probs.csv')
```

morpheus/_lib/src/io/deserializers.cpp
dagardner-nv: done

dagardner-nv:
@mdemoret-nv I addressed your concerns regarding the `make_from_file` as best I can. While I think the method is useful, your concerns a valid, and if you still have issues with it I can remove the method as it is largely unrelated to this fix.

mdemoret-nv (approved):
Looks good now. The roundtrip tests will help keep things stable.

PR #739, opened 2023-03-01
[FEA]: External issue label

jarmak-nv:
This PR automatically adds the https://github.com/nv-morpheus/Morpheus/labels/%3F%20-%20Needs%20Triage label to all issues authored by people who have an [author_association](https://docs.github.com/en/graphql/reference/enums#commentauthorassociation) value of that is not `OWNER` or `MEMBER`. 

The workflow is entirely skipped for org members/owners.

Right now, that means collaborators are labeled, easy to update if we want to change that.


mdemoret-nv (approved):
Looks cool. Eager to try it out.

PR #752, opened 2023-03-09
Add InMemorySinkStage

### LINKED ISSUE #746: [FEA]: Add a `InMemorySinkStage` class to aid in testing

New Feature

Medium

Many of our tests follow this pattern:
```python
    pipe.add_stage(
        WriteToFileStage(config, filename=out_file, overwrite=False, include_index_col=(output_type == "json")))
    pipe.run()

    assert_path_exists(out_file)

    validation_file = os.path.join(TEST_DIRS.tests_data_dir, "filter_probs.csv")
    input_data = np.loadtxt(validation_file, delimiter=",", skiprows=1)

    if output_type == "csv":
        # The output data will contain an additional id column that we will need to slice off
        output_data = np.loadtxt(out_file, delimiter=",", skiprows=1)
    else:  # assume json
        df = read_file_to_df(out_file, file_type=FileTypes.Auto)
        output_data = df.values

    # Somehow 0.7 ends up being 0.7000000000000001
    output_data = np.around(output_data, 2)
    assert output_data.tolist() == input_data.tolist()
```
Which does the following:

1. Writes the pipeline output to a file
2. Checks that the file exists
3. Reads the file
4. Then compares the data

This is slow, cumbersome and error prone for testing when all we need is the final dataframe loaded into memory.

Instead, we should have a stage similar to the `ValidationStage` (Or it could be an improvement to the `ValidationStage`) which holds all messages sent to it in memory. Then once the pipeline is complete, the user could then access the concatenated dataframe. This would function exactly the same as the `WriteToFileStage` and reading the file, except with less steps.

There is already something similar [here](https://github.com/nv-morpheus/Morpheus/blob/branch-23.03/tests/test_nonlinear_pipeline.py#L79) and should be used as a starting point.

One alternative would be to make the `InMemorySinkStage` a base class for several other utility classes. For example, the hierarchy could be:

1. `InMemorySinkStage`
   1. Does nothing more than concat all messages into a single dataframe.
1. `CompareDataframeStage`
   1. Derives from `InMemorySinkStage` and performs `compare_df.compare_df()` saving the results to a member
1. `ValidationStage`
   1. Derives from `CompareDataframeStage` and writes the results calculated in `CompareDataframeStage` to a file like it currently does

### Additional context

_No response_

### Code of Conduct

dagardner-nv:
* Adds two new in-memory stages `InMemorySourceStage` & `InMemorySinkStage`
* `CompareDataFrameStage`, which was formerly a part of the tests, now inherits from `InMemorySinkStage` and is a part of the public API
* `ValidationStage` now inherits from `CompareDataFrameStage`
* Stages which were in `tests/utils.py` have now been moved into their own modules under `tests/stages/`
* The `set_source` and `add_stage` methods of `LinearPipeline` now return the stage being added
* Fix a bug in `compare_df` when `include_columns` is not `None`.
* `tests/test_serialize_stage_pipe.py` folded into `tests/test_file_in_out.py` since it was an end-to-end test and not a true test of the serialize stage.
* Most tests now make use of the new `filter_probs_df` fixture along with the `InMemorySourceStage` & `CompareDataFrameStage` stages rather than `FileSourceStage` & `WriteToFileStage`, this cuts down on the amount of temporary files written to disk and redundant code. 

Note: Tests which either explicitly test either the `FileSourceStage` & `WriteToFileStage` stages or replicate a real-world pipeline remain unchanged.

Fixes #746


tests/stages.py
```diff
@@ -0,0 +1,322 @@
...
+from morpheus.pipeline.preallocator_mixin import PreallocatorMixin
+from morpheus.pipeline.single_output_source import SingleOutputSource
+from morpheus.pipeline.single_port_stage import SinglePortStage
+from morpheus.pipeline.stream_pair import StreamPair
+from morpheus.utils import compare_df
+from morpheus.utils.atomic_integer import AtomicInteger
+
+
+@register_stage("unittest-conv-msg")
+class ConvMsg(SinglePortStage):
```
mdemoret-nv, line 41: Move all of these stages into a subpackage. Add `stages/__init__.py` and then add each stage to a separate file.

tests/test_add_classifications_stage_pipe.py
```diff
@@ -91,19 +86,11 @@ def test_add_classifications_stage_multi_segment_pipe(config, tmp_path):
...
-    assert_path_exists(out_file)
-
-    input_data = np.loadtxt(input_file, delimiter=",", skiprows=1)
-    expected = (input_data > threshold)
-
-    # The output data will contain an additional id column that we will need to slice off
-    output_data = pd.read_csv(out_file)
-    idx = output_data.columns.intersection(config.class_labels)
+    results_df = comp_stage.concat_dataframes()
+    idx = results_df.columns.intersection(config.class_labels)
```
mdemoret-nv, line 93: Change this check to use the output of `get_results()`

mdemoret-nv (approved):
Excited for this change

mdemoret-nv (approved):

PR #754, opened 2023-03-10
Improve the python CMake functionality to speed up the configure step

mdemoret-nv:
This is the corresponding update to incorporate changes from https://github.com/nv-morpheus/utilities/pull/17.

Summary of changes:

1. This reorganizes the python CMake checks so they only get executed once (handled in utilities). Significantly speeding up the configure step.
1. Makes the CMake variable names consistent (breaking change)
   1. For example, `MORPHEUS_BUILD_PYTHON_STUBS` -> `MORPHEUS_PYTHON_BUILD_STUBS`
1. Removes extra cmake code that is now handled by utilities


CMakeLists.txt
```diff
@@ -116,15 +118,17 @@ if("${CMAKE_EXECUTABLE_SUFFIX}" STREQUAL "")
   set(CMAKE_EXECUTABLE_SUFFIX ".x")
 endif()
 
+# Create a custom target to allow preparing for style checks
```
cwharris, line 121: how come we're moving this above the python config? we just don't need it?

morpheus/_lib/cmake/libraries/cuda_utils.cmake
```diff
@@ -35,9 +35,6 @@ set_target_properties(
 target_include_directories(cuda_utils_objs
     PUBLIC
       "${MORPHEUS_LIB_ROOT}/include"
-      cudf::cudf
```
cwharris, line 38: how come we can get rid of these now?

cwharris (approved):

morpheus/_lib/cmake/libraries/cuda_utils.cmake
mdemoret-nv: These are transitively brought in from the `cudf_helpers` library so this is not necessary. Also we shouldnt be adding includes like this (we should use `target_link_libraries`)

CMakeLists.txt
mdemoret-nv: Oh, I just moved it above the big block comment `Morpheus Python Setup` so non-python stuff is before and python stuff is after. It doesnt matter the order of this an `morpheus_utils_python_configure()`

PR #635, opened 2023-01-26
Add md & rst files to copyright.py checks and Fix copyright year in Sphinx footer

dagardner-nv:
We now have a significant amount of content in markdown and restructured text files, we should start ensuring they have copyright headers.

PR #759, opened 2023-03-13
[FEA] Remove md and rst from copyright check

jarmak-nv:
Minor change to the copyright CI check - text documents (`.md` and `.rst`) presented as-is to users don't need a copyright header.

### DIFF
--- a/ci/scripts/copyright.py
+++ b/ci/scripts/copyright.py
@@ -30,7 +30,7 @@ import gitutils  # noqa: E402
 
 FilesToCheck = [
     # Get all of these extensions and templates (*.in)
-    re.compile(r"[.](cmake|cpp|cc|cu|cuh|h|hpp|md|rst|sh|pxd|py|pyx|yml|yaml)(\.in)?$"),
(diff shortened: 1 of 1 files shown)


dagardner-nv (commented):
Enforcing copyright headers on md & rst files was explicitly added in https://github.com/nv-morpheus/Morpheus/pull/635 the reasoning being that we had a significant amount of content in our documentation.

jarmak-nv:
Fair enough! I'll update this to exclude specifically the `PULL_REQUEST_TEMPLATE.md` file

dagardner-nv:
> Fair enough! I'll update this to exclude specifically the `PULL_REQUEST_TEMPLATE.md` file

You could also just ignore anything in the `.github` dir, or just toss the license header in there if we wrote it.

jarmak-nv:
> You could also just ignore anything in the `.github` dir, or just toss the license header in there if we wrote it.

Yeah I thought about the `.github` dir but I think some of the actions we write/use might need it in the future so leaving it in. I would just toss the license in but I don't know for sure if we're the original authors and people would need to delete it each time they opened a PR.

Also after talking with @mdemoret-nv I can just update `copyright.py` in the PR that adds the new file, so closing this.

Issue #331, opened 2022-08-22
[FEA] Allow users to write, compile, and load custom sources/stages/sinks without recompiling all of morpheus


dagardner-nv:
**Is your feature request related to a problem? Please describe.**
Currently, users need to build Morpheus from source in order to get access to the CMake configuration and header files necessary to develop C++ stages & messages outside of the source tree.

**Describe the solution you'd like**
Install headers & cmake config file. This requires altering the CMake files to add the correct `install` options.

mdemoret-nv:
### Completion Criteria

- [ ] Updated CMake files to include C++ headers and CMake configuration files when installing the CMake project
- [ ] Inclusion of the headers and config files in the Morpheus conda package (Should be included automatically from previous item)
- [ ] Addition of an example which builds a C++ stage using Morpheus headers
   - [ ] This example should be optionally included when building from the Morpheus root (See MRC for an example)
   - [ ] This example should be able to be compiled from outside the Morpheus root project as well (See MRC for an example)
   - [ ] This project should install into the python environment to make testing easy (See MRC for an example)
   - [ ] This example should be built as a part of CI to ensure it does not get out of date
- [ ] [Optional] Update developer examples to follow the pattern established in the example above

PR #763, opened 2023-03-15
Migrate dfencoder to morpheus repo

### LINKED ISSUE #753: Migrate `dfencoder` into the Morpheus repo
To begin the process outlined in #517, `dfencoder` should be moved into the morpheus repo. This will involve:

- Moving all files with as little modification as possible (to allow any existing PRs like https://github.com/nv-morpheus/dfencoder/pull/11 to be easily migrated)
   - Imports should be the only things that need to be updated
- Moved into a new subpackage: `morpheus.models.dfencoder` (open to a better location)
- `dfencoder` removed from the dependencies of Morpheus
- A basic testing framework put in place
   - Realistically, we just need a single test that imports the model and does a simple training run with random data
   - This is to ensure that all of the necessary dependencies are installed
- The existing repo set to archived mode

### LINKED ISSUE #765: [BUG]: C++ tests not being run in CI
### Version

23.03

Source

### Describe the bug.

We currently only transfer the wheel, this excludes `./build/morpheus/_lib/tests/test_libmorpheus.x`

In the test stage we do:
```
CPP_TESTS=($(find ${MORPHEUS_ROOT}/build -name "*.x"))
```

but that just creates an empty bash array

_No response_

### Relevant log output

_No response_

### Full env printout

_No response_

### Other/Misc.

_No response_

### Code of Conduct

dagardner-nv:
* Copy dfencoder code, ensuring correct copyright headers
* Add a `manual_seed` helper method, replacing duplicated code
* Add tests for dfencoder
* Insert both our Apache license, and the original author's BSD license in each of the 4 python files (and the `__init__.py`). The original code didn't have that, but this conforms with our other 3rd party code inclusions.
* Transfer compiled C++ tests and libs (fixes #765)

fixes #753


morpheus/models/dfencoder/autoencoder.py
```diff
@@ -0,0 +1,1129 @@
...
+    and optional activation and dropout."""
+
+    def __init__(self, in_dim, out_dim, activation=None, dropout=None, *args, **kwargs):
+        super(CompleteLayer, self).__init__(*args, **kwargs)
+        self.layers = []
+        linear = torch.nn.Linear(in_dim, out_dim)
+        self.layers.append(linear)
+        self.add_module('linear_layer', linear)
+
+        self.activation = activation
```
dagardner-nv, line 101: This is a change I added, fixes an obvious bug in the `interpret_activation` method when `act=None`

tests/conftest.py
```diff
@@ -28,6 +28,8 @@
 import pytest
 import requests
 
+import cudf
+
```
mdemoret-nv, line 32: We really dont want this here if possible. It will slow everything down including just getting the tests name. Do a local import where its needed.

tests/conftest.py
```diff
@@ -344,6 +375,13 @@ def reload_modules(request: pytest.FixtureRequest):
             importlib.reload(mod)
 
 
+@pytest.fixture(scope="function")
+def manual_seed():
+    from morpheus.utils import seed as seed_utils
+    seed_utils.manual_seed(42)
+    yield
+
+
```
mdemoret-nv, line 384: Nice function. Couple of things:

1. Needs a docstring to explain what its being used for
2. This could also return a function to allow users to call it within the test. Like `yield lambda _: seed_utils.manual_seed(42)`

tests/conftest.py
```diff
@@ -354,6 +363,25 @@ def chdir_tmpdir(request: pytest.FixtureRequest, tmp_path):
...
+@pytest.fixture(scope="function")
+def df(_filter_probs_df: cudf.DataFrame, df_type: typing.Literal['cudf', 'pandas'], use_cpp: bool):
+    if df_type == 'cudf':
+        yield _filter_probs_df.copy(deep=True)
+    elif df_type == 'pandas':
+        yield _filter_probs_df.to_pandas()
+    else:
+        assert False, "Unknown df_type type"
+
+
```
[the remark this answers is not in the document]
mdemoret-nv: Created issue: https://github.com/nv-morpheus/Morpheus/issues/785

tests/test_filter_detections_stage.py
```diff
@@ -14,20 +14,24 @@
...
 from morpheus.stages.postprocess.filter_detections_stage import FilterDetectionsStage
-from utils import TEST_DIRS
+
+
+@pytest.fixture
+def df_type():
+    """
+    The tests in this module all require a cudf dataframe
+    """
+    yield 'cudf'
```
mdemoret-nv, line 34: Mark tests with `use_cudf` if they should only be run with cuDF (I just added this feature)

tests/test_filter_detections_stage.py
```diff
@@ -112,32 +113,26 @@ def test_filter_copy(config):
...
-    input_file = os.path.join(TEST_DIRS.tests_data_dir, "filter_probs.csv")
-    df = read_file_to_df(input_file, file_type=FileTypes.Auto, df_type='cudf')
-
+def test_filter_column(config, df, do_copy, threshold, field_name):
     fds = FilterDetectionsStage(config,
                                 threshold=threshold,
                                 copy=do_copy,
                                 filter_source=FilterSource.DATAFRAME,
                                 field_name=field_name)
+    expected_df = df.to_pandas()
```
mdemoret-nv, line 122: Use marks instead of converting

tests/test_filter_detections_stage_column_pipe.py
```diff
@@ -32,12 +30,20 @@
 from utils import ConvMsg
 
 
+@pytest.fixture
+def df_type():
+    """
+    We only need the df for our expected valus, and we don't want to double the number of our tests.
+    """
+    yield 'pandas'
```
mdemoret-nv, line 38: Use marks again here.

tests/conftest.py
```diff
@@ -354,6 +392,25 @@ def chdir_tmpdir(request: pytest.FixtureRequest, tmp_path):
...
+def _filter_probs_df():
+    from morpheus._lib.common import FileTypes
+    from morpheus.io.deserializers import read_file_to_df
+    from utils import TEST_DIRS
+    input_file = os.path.join(TEST_DIRS.tests_data_dir, "filter_probs.csv")
+    yield read_file_to_df(input_file, file_type=FileTypes.Auto, df_type='cudf')
+
+
+@pytest.fixture(scope="function")
+def df(_filter_probs_df: cudf.DataFrame, df_type: typing.Literal['cudf', 'pandas'], use_cpp: bool):
```
mdemoret-nv, line 405: `df_type` should be on the `_filter_probs_df` fixture.

tests/test_message_meta.py
```diff
@@ -36,25 +32,29 @@ def index_type(request: pytest.FixtureRequest) -> typing.Literal["normal", "skip
...
 
-    loaded_df = read_file_to_df(os.path.join(TEST_DIRS.tests_data_dir, 'filter_probs.csv'),
-                                file_type=FileTypes.Auto,
-                                df_type=df_type)
+    if df_type == 'cudf':
+        loaded_df = _filter_probs_df.copy(deep=True)
+    elif df_type == 'pandas':
+        loaded_df = _filter_probs_df.to_pandas()
+    else:
+        assert False, "Unknown df_type type"
```
mdemoret-nv, line 45: This should be changed to just `filter_probs_df` which should be parameterized already. Then we dont need any of this conversion code to be repeated

tests/test_message_meta.py
```diff
@@ -36,25 +32,29 @@ def index_type(request: pytest.FixtureRequest) -> typing.Literal["normal", "skip
...
-        return loaded_df.iloc[::3, :].copy()
+        yield loaded_df.iloc[::3, :].copy()
     elif (index_type == "dup"):
         # Duplicate
-        return duplicate_df_index_rand(loaded_df, count=2)
+        yield duplicate_df_index_rand(loaded_df, count=2)
     elif (index_type == "down"):
         # Reverse
-        return loaded_df.iloc[::-1, :].copy()
+        yield loaded_df.iloc[::-1, :].copy()
```
mdemoret-nv, line 57: Why were these switched to yield instead of return? You can still assert with returns

tests/test_nonlinear_pipeline.py
```diff
@@ -37,6 +36,14 @@
 from utils import TEST_DIRS
 
 
+@pytest.fixture
+def df_type():
+    """
+    We only need the df for our expected valus, and we don't want to double the number of our tests.
+    """
+    yield 'pandas'
```
mdemoret-nv, line 44: Use marks here

mdemoret-nv (approved):
Tests for dfencoder look great. Exactly what we need for unit tests. Quite a few comments but I'm approving since they are all straightforward.

setup.cfg
```diff
@@ -80,6 +80,7 @@ exclude =
    cpp,
    docs,
    models/,
+   morpheus/models/dfencoder/*.py,
```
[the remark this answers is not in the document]
dagardner-nv: Done https://github.com/nv-morpheus/Morpheus/issues/786

tests/test_message_meta.py
dagardner-nv: The reason why it isn't is that the `_filter_probs_df` fixture is a session scoped fixture, while `filter_probs_df` is a per-function fixture. 

So that we read it once, and make a per-function copy.

ci/scripts/github/build.sh
```diff
@@ -30,11 +30,14 @@ sccache --version
...
 git submodule update --init --recursive
+
+# Explicitly setting nvcc path Work-around for https://github.com/rapidsai/cudf/issues/12862
 cmake -B build -G Ninja ${CMAKE_BUILD_ALL_FEATURES} \
     -DCCACHE_PROGRAM_PATH=$(which sccache) \
     -DMORPHEUS_PYTHON_BUILD_WHEEL=ON \
     -DMORPHEUS_PYTHON_BUILD_STUBS=OFF \
-    -DCMAKE_BUILD_RPATH_USE_ORIGIN=ON .
+    -DCMAKE_BUILD_RPATH_USE_ORIGIN=ON \
+    -DCMAKE_CUDA_COMPILER=${CUDA_PATH}/bin/nvcc  .
```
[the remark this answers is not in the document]
mdemoret-nv: See PR #764

tests/test_message_meta.py
mdemoret-nv: Understood. We should have a session scoped `_filter_probs_df` and then a function scoped `filter_probs_df` fixture. The `filter_probs_df` should request both `df_type` and `use_cpp` so it gets parameterized on both. Then in `test_message_meta` all we need to do is add the `filter_probs_df` fixture which will handle this code for us.

tests/test_message_meta.py
mdemoret-nv: Yes, but there is no cleanup code here. One is not preferable over the other. Just dependent on the situation.

tests/test_filter_detections_stage.py
dagardner-nv: For this test I need the input to be cudf, but I'm using pandas to build out the expected results.

ci/scripts/github/build.sh
```diff
@@ -45,8 +48,16 @@ sccache --show-stats
...
+MORPHEUS_LIBS=($(find ${MORPHEUS_ROOT}/build/morpheus/_lib -name "*.so" -exec realpath --relative-to ${MORPHEUS_ROOT} {} \;))
+tar cfj "${WORKSPACE_TMP}/morhpeus_libs.tar.bz" "${MORPHEUS_LIBS[@]}"
+
+CPP_TESTS=($(find ${MORPHEUS_ROOT}/build/morpheus/_lib/tests -name "*.x" -exec realpath --relative-to ${MORPHEUS_ROOT} {} \;))
+tar cfj "${WORKSPACE_TMP}/cpp_tests.tar.bz" "${CPP_TESTS[@]}"
+
 rapids-logger "Pushing results to ${DISPLAY_ARTIFACT_URL}"
 aws s3 cp --no-progress "${WORKSPACE_TMP}/wheel.tar.bz" "${ARTIFACT_URL}/wheel.tar.bz"
+aws s3 cp --no-progress "${WORKSPACE_TMP}/morhpeus_libs.tar.bz" "${ARTIFACT_URL}/morhpeus_libs.tar.bz"
+aws s3 cp --no-progress "${WORKSPACE_TMP}/cpp_tests.tar.bz" "${ARTIFACT_URL}/cpp_tests.tar.bz"
```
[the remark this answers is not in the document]
dagardner-nv: Copying the C++ tests (currently just `./build/morpheus/_lib/tests/test_libmorpheus.x`) and the `*.so` files in the build dir that `test_libmorpheus.x` is linked against. 

Since we don't install `test_libmorpheus.x` the rpath isn't updated. In theory we could avoid copying the `*.so` files since `conda/envs/morpheus/lib` is in the rpath, but that would require separating the morpheus installs the way we do for MRC (#331 should do this).

ci/scripts/github/build.sh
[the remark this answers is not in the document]
dagardner-nv: I ran into a ton of trouble on this one, some of which is documented in https://github.com/rapidsai/cudf/issues/12862 

The issue is without specifying `CMAKE_CUDA_COMPILER` cmake will pick up `nvcc` from the conda env, which should be OK, but the problem is something in the stack (I'm not sure exactly what) is using the location of `nvcc` to infer the location of the CUDA libs, causing `libmorpheus.so` to link against the sub libs in the conda env. 

Without this calling `pybind11::module_::import("cudf")` or `CudfHelper::load()` causes a Runtime error because `cuInit` is undefined. However this only happens when Python is being used in embedded mode (ie. the C++ tests).

ci/runner/Dockerfile
```diff
@@ -58,6 +58,7 @@ ARG CUDA_PKG_VER
 RUN apt update && \
     DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC \
     apt install --no-install-recommends -y \
+    build-essential \
```
mdemoret-nv, line 61: Do we need this anymore?

ci/scripts/github/build.sh
```diff
@@ -30,6 +30,8 @@ sccache --version
 
 rapids-logger "Configuring cmake for Morpheus"
 git submodule update --init --recursive
+
+# Explicitly setting nvcc path Work-around for https://github.com/rapidsai/cudf/issues/12862
```
mdemoret-nv, line 34: Comment isnt needed anymore

mdemoret-nv (approved):

PR #758, opened 2023-03-13
Document adding cuML and fix the gnn_fraud_detection_pipeline example

dagardner-nv:
* Install stellargraph and tensorflow via pip, the versions in conda-forge are tied to cuda-11.3
* Use tensorflow-2.12, as prior versions conflicted with our version of protobuf.
* Explicitly install a compatable version of protobuf from the anaconda channel that is not in the conda-forge channel.
* Work-around the Keras model's `predict` method no longer accepting the return value from `HinSAGENodeGenerator.flow` due to switch to tensorflow-2.12

fixes #741

PR #764, opened 2023-03-15
Have conda-build test cuml installation

cwharris:
contributes to https://github.com/nv-morpheus/Morpheus/issues/706, as the base container for devcontainer will also be used for the ci container.


cwharris:
@dagardner-nv in https://github.com/nv-morpheus/Morpheus/pull/758 you added cuml to the development environment. since then, @mdemoret-nv mentioned installing cuml as part of the test environment rather than the development environment, which I've done in this pr. Any thoughts on this?

dagardner-nv:
> @dagardner-nv in #758 you added cuml to the development environment. since then, @mdemoret-nv mentioned installing cuml as part of the test environment rather than the development environment, which I've done in this pr. Any thoughts on this?

I need to rework my PR a bit. We don't need cuml directly in morpheus, but we do need to pin some of our deps such that users are able to install cuml after the fact without breaking the env. I like the idea of having it in the test container.

dagardner-nv (approved):

docker/Dockerfile
```diff
@@ -146,6 +146,7 @@ RUN apt update && \
         cuda-cupti-dev-${CUDA_MAJOR_VER}-${CUDA_MINOR_VER} \
         cuda-nvml-dev-${CUDA_MAJOR_VER}-${CUDA_MINOR_VER} \
         cuda-nvtx-${CUDA_MAJOR_VER}-${CUDA_MINOR_VER} \
+        cuda-compiler-${CUDA_MAJOR_VER}-${CUDA_MINOR_VER} \
```
dagardner-nv, line 149: Minor issue, but lets keep the packages sorted

mdemoret-nv (approved):

PR #776, opened 2023-03-20
Fix for TritonInferenceStage memory leak

cwharris:
Fixes a memory leak in `TritonInferenceStage`, caused by a un-deleted pointer to inference results.


dagardner-nv (approved):
Nice catch

PR #846, opened 2023-04-05
Add DataManager object to morpheus/io

### LINKED ISSUE #845: [FEA]: Add new IO primitive to manage in memory or on disk dataframe storage

New Feature

Critical (currently preventing usage)

Add a new primitive that will allow collections of DataFrames to be serialized to system memory or disk and . This will allow for more seamless handling of training and inference datasets going forward.

I'm going to add a new DataManager class to `morpheus/io/data_manager.py`.

This class will be fairly light weight and leverage fsspec where possible.

There are various features of Dask and Arrow that we might incorporate later, but I'm not aware of anything that handles the specific problem I'm trying to address here.

### Additional context

_No response_

### Code of Conduct

drobison00:
Closes #845


morpheus/io/data_manager.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def _update_manifest(self, source_id: uuid.UUID, action: str) -> None:
+        data_record = self._records[source_id]
+
+        if action == 'store':
+            self._manifest[source_id] = data_record.data
+            self._total_rows += data_record.num_rows
+        elif action == 'remove':
+            self._total_rows -= data_record.num_rows
+            del self._manifest[source_id]
```
mdemoret-nv, line 110: Why have a separate manifest? Wouldnt it be easier to generate this on demand from the `_records` when its requested? Same with `_total_rows`?

morpheus/io/data_record.py
```diff
@@ -0,0 +1,275 @@
...
+        elif self._file_format == 'parquet':
+            self._data_reader = cudf.read_parquet
+
+        self._store(data_source)
+
+    def __del__(self):
+        """Delete the DataRecord instance."""
+        try:
+            if (self._owner and self._storage_type == 'filesystem'):
+                self._fs.rm(self._backing_source)
```
mdemoret-nv, line 81: Things like this lead me to believe that a better implementation would be one with `self._data` point to an abstract class `DataRecordStorage` which handles all of the actually reading/writing, deletion, etc. Then you have different implementations for in memory, filesystem, etc. And you could even make utility classes that implement the interface but act as a cache and fallback through chained implementations in case of a cache miss.

Keeping pairs of properties like this working correctly will be difficult as the number of storage types increases.

mdemoret-nv (changes requested):
We should discuss some more concrete implementations outside of DFP to better understand how this will be used. Ideally, we are forward thinking in the API, even if the implementation isnt complete. That way we can fill out the implementation over time without breaking changes.

tests/io/data_storage/test_file_system_storage.py
```diff
@@ -0,0 +1,151 @@
...
+def csv_file(csv_data):
+    csv_data.to_csv('test_data.csv', index=False)
+    yield 'test_data.csv'
+    os.remove('test_data.csv')
+
+
+@pytest.fixture
+def parquet_file(parquet_data):
+    parquet_data.to_parquet('test_data.parquet', index=False)
+    yield 'test_data.parquet'
```
mdemoret-nv, line 46: We should have these write to temporary file storage.

tests/io/test_data_manager.py
```diff
@@ -58,14 +59,14 @@ def dataframe_fixture_data():
     test_cudf_dataframe.to_csv(csv_filepath, index=False, header=True)
```
mdemoret-nv, line 59: Lets use the temp file fixture here. Also remove `print` statements from tests

mdemoret-nv (approved):

PR #848, opened 2023-04-05
Update to CuDF 23.02 and MRC 23.07

cwharris:


ci/conda/recipes/morpheus/meta.yaml
```diff
@@ -72,8 +72,8 @@ outputs:
         # Runtime only requirements. This + setup.py is the definitive runtime requirement list
         - click >=8
         - configargparse 1.5.*
-        - cuda-python <=11.7.0 # Needed to ensure cudatoolkit and not cuda-toolkit is installed
-        - cudf
+        - cuda-python 11.8
+        - cudf {{ rapids_version }}.*
```
mdemoret-nv, line 76: If you specify a dependency in the `host` section, you shouldnt put version requirements in the `run` section unless you need a specific version that is more restrictive than what is specified in `host`. This allows the dependency itself to specify its ABI compatibility.

ci/conda/recipes/morpheus/meta.yaml
```diff
@@ -55,7 +55,7 @@ outputs:
         - ccache
         - ninja
       host:
-        - cuda-python <=11.7.0
+        - cuda-python 11.8
```
mdemoret-nv, line 58: If we go up to `cuda-python>11.7` then we should specify the version requirements on `cuda-toolkit` since this will now get brought it too.

scripts/compile.sh
```diff
@@ -27,6 +27,7 @@ cmake -B ${BUILD_DIR} -GNinja \
    -DMORPHEUS_PYTHON_INPLACE_BUILD=ON \
    -DMORPHEUS_USE_CCACHE=ON \
    -DMORPHEUS_USE_CONDA=${MORPHEUS_USE_CONDA:-"ON"} \
+   -DMRC_VERSION="23.07" \
```
mdemoret-nv, line 30: This shouldnt be needed.

mdemoret-nv (commented):
Couple of comments.

ci/conda/recipes/morpheus/meta.yaml
cwharris: We don't have a direct dependency on `cuda-toolkit`, why is it important that we include it here? That seems like a problem if cuda-python gets more specific about the `cuda-*` packages it depends on.
        - cuda-python 11.8
```

File: morpheus/_lib/cudf_helpers.pyx [hunk 4]
```diff
@@ -58,17 +58,17 @@ cdef public api:
...
                 for i in range(min(index_col_count, table.metadata.schema_info.size())):
                     index_names.append(table.metadata.schema_info[i].name.decode())
 
         column_names = []
 
         # Need to support both column_names and schema_info
-        if (table.metadata.column_names.size() > 0):
-            column_names = [x.decode() for x in table.metadata.column_names[index_col_count:]]
+        if (table.metadata.schema_info.size() > 0):
+            column_names = [x.name.decode() for x in table.metadata.schema_info][index_col_count:]
```
mdemoret-nv, line 71: Lets combine the conversion to python strings enumeration into one array and then slice for index and column names separately

mdemoret-nv (approved):

docker/conda/environments/cuda11.8_dev.yml
```diff
@@ -31,11 +31,12 @@ dependencies:
...
     - cmake=3.24
     - configargparse=1.5
+    - cuda-compiler=11.8
     - cuda-nvml-dev=11.8
-    - cuda-python<=11.7.0 # Remove when Issue #251 is closed
     - cudatoolkit=11.8
-    - cudf 22.10
+    - cudf 23.02
     - cupy=9.5.0
+    - cxx-compiler
```
mdemoret-nv, line 39: With the addition of this package, it automatically brings in `gxx_linux-64` and `gcc_linux-64`.
```
$ mamba repoquery depends cxx-compiler

Executing the query cxx-compiler



 Name         Version Build       Channel    
──────────────────────────────────────────────
 cxx-compiler 1.5.2   hf52228f_0  conda-forge
 c-compiler   1.5.2   h0b41bf4_0  conda-forge
 gxx          11.2.0  h702ea55_10 conda-forge
 gxx_linux-64 11.2.0  hacbe6df_10 conda-forge
```
Should we remove them from our dependencies?

ci/conda/recipes/morpheus/conda_build_config.yaml
```diff
@@ -33,5 +33,5 @@ boost:
   - 1.74
 
 rapids_version:
-  - 22.08
   - 22.10
+  - 23.02
```
mdemoret-nv, line 37:
```suggestion
  - 23.02
```

For morpheus, we should reduce this to just one version of rapids.

docker/conda/environments/cuda11.8_dev.yml
```diff
@@ -31,11 +31,12 @@ dependencies:
     - click >=8
     - cmake=3.24
     - configargparse=1.5
+    - cuda-compiler=11.8
     - cuda-nvml-dev=11.8
```
mdemoret-nv, line 35: If the channel `nvidia/label/cuda-11.8.0` is before `nvidia`, can we remove the version from all of the `cuda-*` packages? That way we can reduce the number of places that you specify the cuda version.

mdemoret-nv (changes requested):
Looks good. Thought of a few more things. Also, can you look at whether [this](https://github.com/nv-morpheus/Morpheus/blob/branch-23.07/morpheus/_lib/src/io/deserializers.cpp#L76-L105) is needed anymore? We used to have to perform some cleanup when reading JSON with cudf. With their new JSON deserializer, I wonder if this is needed anymore. Matching code for python is located [here](https://github.com/nv-morpheus/Morpheus/blob/branch-23.07/morpheus/io/deserializers.py#L29)

docker/conda/environments/cuda11.8_dev.yml
cwharris: possibly. I'll check now that it's in place.

cwharris: no, because what we really want to do is move conda-forge before any nvidia channels to better fit with the rapids ecosystem.

mdemoret-nv (approved):
Looks good once tests pass

PR #877, opened 2023-04-13
Removing explicit driver install from CI runner

mdemoret-nv:
## Description
This PR is follows the MRC PR: https://github.com/nv-morpheus/MRC/pull/313 which removes the explicit dependency on `libnvidia-ml.so` which allows us to no longer need the driver installed in our CI runner.

This allows us to use the `gpu-v100-latest-1` tag to stay up to date with the CI images.

@jjacobelli for Viz


dagardner-nv (approved):
I like it!

Issue #859, opened 2023-04-07
[BUG]: cupy, and by extension cudf changes the system's preferred encoding to ANSI_X3.4-1968


dagardner-nv:
### Version

23.07

### Which installation method(s) does this occur on?

Docker, Conda, Source

### Describe the bug.

Calling various cupy and cudf methods changes the system's preferred encoding from UTF-8 to ANSI_X3.4-1968.
https://github.com/cupy/cupy/issues/7514
https://github.com/rapidsai/cudf/issues/13085

The problem is any code called after this that requires reading a UTF-8 data source without explicitly setting the encoding will fail.

### Minimum reproducible example

```shell
import locale

import cupy as cp

print(locale.getpreferredencoding()) # UTF-8
cpa =  cp.arange(0, 10)
print(locale.getpreferredencoding()) # ANSI_X3.4-1968

with open('models/training-tuning-scripts/sid-models/resources/bert-base-cased-vocab.txt') as fh: # contains unicode chars
    contents = fh.read()

```

### Relevant log output

```shell
UTF-8
ANSI_X3.4-1968

---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
Cell In[1], line 10
      7 print(locale.getpreferredencoding()) # ANSI_X3.4-1968
      9 with open('models/training-tuning-scripts/sid-models/resources/bert-base-cased-vocab.txt') as fh: # contains unicode chars
---> 10     contents = fh.read()

File ~/work/conda/envs/morpheus/lib/python3.8/encodings/ascii.py:26, in IncrementalDecoder.decode(self, input, final)
     25 def decode(self, input, final=False):
---> 26     return codecs.ascii_decode(input, self.errors)[0]

UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 1323: ordinal not in range(128)
```

### Code of Conduct

- [X] I agree to follow Morpheus' Code of Conduct
- [X] I have searched the [open bugs](https://github.com/nv-morpheus/Morpheus/issues?q=is%3Aopen+is%3Aissue+label%3Abug) and have found no duplicates for this bug report

dagardner-nv:
This is a known issue with NVRTC (https://github.com/cupy/cupy/issues/7514#issuecomment-1500656392) currently there are only two known work-arounds:

1. Setting `LC_ALL="POSIX"` 
2. Explicitly set the encoding when opening a file ex: `with open(vocab_path, encoding='UTF-8')`

Issue #849, opened 2023-04-05
[FEA]: Create tests for custom stages in examples


dagardner-nv:
### Is this a new feature, an improvement, or a change to existing functionality?

Improvement

### How would you describe the priority of this feature request

Medium

### Please provide a clear description of problem this feature solves

These are frequently broken as we currently don't have any test coverage for these.

### Describe your ideal solution

Write tests

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I have searched the [open feature requests](https://github.com/nv-morpheus/Morpheus/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22%2Cimprovement%2Cenhancement) and have found no duplicates for this feature request

dagardner-nv:
The last of this work was fixed by PR #940

PR #885, opened 2023-04-13
Create tests for examples with custom stages

### LINKED ISSUE #850: [FEA]: Remove redundant bert hash files

Change

Low (would be nice)

Looks like we have a couple of different copies of these, not sure if git-lfs is smart enough to prevent redundant copies of these, but even if it is, we have these also stored in the `morpheus/data` dir as regular git files.

```
$ find ./models/ -name "bert-base-cased-hash.txt" -exec diff -s morpheus/data/bert-base-cased-hash.txt {} \;
Files morpheus/data/bert-base-cased-hash.txt and ./models/training-tuning-scripts/sid-models/resources/bert-base-cased-hash.txt are identical
Files morpheus/data/bert-base-cased-hash.txt and ./models/training-tuning-scripts/log-parsing-models/resources/bert-base-cased-hash.txt are identical
$ find ./models/ -name "bert-base-uncased-hash.txt" -exec diff -s morpheus/data/bert-base-uncased-hash.txt {} \;
Files morpheus/data/bert-base-uncased-hash.txt and ./models/training-tuning-scripts/phishing-models/resources/bert-base-uncased-hash.txt are identical
Files morpheus/data/bert-base-uncased-hash.txt and ./models/training-tuning-scripts/sid-models/resources/bert-base-uncased-hash.txt are identical
Files morpheus/data/bert-base-uncased-hash.txt and ./models/training-tuning-scripts/root-cause-models/resources/bert-base-uncased-hash.txt are identical
```

delete

_No response_

### Additional context

_No response_

### Code of Conduct

dagardner-nv:
## Description
This PR creates at least one test for each example containing custom stages.
This PR currently only covers those examples which do not require additional packages. 
Part of #849.

* Moves the bert vocabulary files to `morpheus/data` dir, no longer requiring them to be fetched from LFS and making them available to unittests.
* Fixes type hints and remove a redundant method in `examples/log_parsing/inference.py`
* Remove redundant copies of `bert-base-cased-hash.txt` and `bert-base-uncased-hash.txt` files, replacing them with symlinks to the files in the morpheus/data` dir fixes #850
* Explicitly set `encoding='UTF-8'` in `examples/log_parsing/postprocessing.py` as a work-around for issue #859 
* Add `py::kw_only` to Python bindings for `TensorMemory` and sublasses to ensure parity with Python impls.
* Set `repr=False` for the `tensors` field of `TensorMemory` avoids bug when printing due to the fact that we assign the value to `self._tensors`
* Seed cupy's random number generator in `manual_seed` method.
* Fix usage of `reload_modules` fixture, requesting a reload of multiple modules should be done with `@pytest.mark.reload_modules([mod1, mod2])` not calling `reload_modules` twice.
* New test data in `tests/tests_data/log_parsing` is based upon the first 5 rows of data from `models/datasets/validation-data/log-parsing-validation-data-input.csv`

## Checklist


tests/examples/log_parsing/conftest.py
```diff
@@ -0,0 +1,23 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+
+
+@pytest.fixture
+def config(config):
```
mdemoret-nv, line 20: Comment on why this is necessary

tests/examples/test_pass_thru.py
```diff
@@ -0,0 +1,55 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
mdemoret-nv, line 1: Move this file into a subdirectory under `tests/examples` to mirror the example it is testing (`developer_guide`)

tests/examples/test_recipient_features_stage.py
```diff
@@ -0,0 +1,63 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
mdemoret-nv, line 1: Move this file into a subdirectory under `tests/examples` to mirror the example it is testing (`developer_guide`)

tests/examples/test_abp_pcap_preprocessing.py
```diff
@@ -24,12 +23,13 @@
 import cudf
 
 from morpheus.common import TypeId
+from morpheus.config import Config
```
mdemoret-nv, line 26: Move this file into a subdirectory under `tests/examples` to mirror the example it is testing (`abp_pcap_detection`)

tests/utils/dataset_manager.py
```diff
@@ -64,15 +64,22 @@ def clear(self):
...
         relative to the `test/tests_data` dir. If a DataFrame matching both `file_path` and `df_type` has already been
         fetched, then a cached copy will be returned. In the event that a DataFrame matching `file_path` but not
         `df_type` exists in the cache, then the cached copy will be cast to the appropriate type, stored in the cache
         and then returned
+
+        Passing values to `reader_kwargs` will cause the cache to by bypassed
         """
 
+        if len(reader_kwargs) and not no_cache:
+            no_cache = True
```
mdemoret-nv, line 81: Generate a warning in this situation

tests/utils/dataset_manager.py
```diff
@@ -83,20 +90,27 @@ def get_df(self,
...
                     df = alt_df.to_pandas()
                 else:
                     df = cdf.DataFrame.from_pandas(alt_df)
             else:
-                df = read_file_to_df(full_path, df_type=df_type)
+                df = read_file_to_df(full_path, df_type=df_type, **reader_kwargs)
 
-            self.__df_cache[(df_type, full_path)] = df
+            if not no_cache:
+                self.__df_cache[(df_type, full_path)] = df
```
mdemoret-nv, line 113: Philosophical question: Does using `no_cache` prevent reading from the cache, writing to the cache, or both? I could argue that its reading only and there is little harm to writing to the cache with `no_cache` set.

mdemoret-nv (changes requested):
A few suggestions:

- Make the `tests/examples` directory better mirror the `examples` directory
   - Since there are no python files in the root of `examples`, there shouldnt be any tests in the root of `tests/examples`
   - Move any of the `tests/examples/test_XXX.py` file in the root of `tests/examples` into a folder to match `examples/XXX`
   - i.e. `tests/examples/test_pass_thru.py` -> `tests/examples/developer_guide/test_pass_thru.py`
- Make the `tests/tests_data/examples` directory better mirror the `examples` directory
   - Same as the previous comment. Things like `tests/tests_data/examples/abp_pcap.jsonlines` should be moved to `tests/tests_data/examples/abp_pcap_detection/abp_pcap.jsonlines`

tests/utils/dataset_manager.py
dagardner-nv: The problem with writing to the cache when `no_cache=True` is set happens when its combined with setting reader_args. 

For example:
```
df = dataset.get_df(input_file, no_cache=True, filter_nulls=False)
```

If `input_file` was cached, a second reader wishing the default behavior of `filter_nulls=True` would get a df with nulls.

mdemoret-nv: If you ignore `reader_args`, do you still feel the same way? I understand that using `reader_args` would completely invalidate using the cache at all. But if the user supplied that argument, what would happen?

dagardner-nv: Potentially we could write to the cache when `reader_args` is empty, but I would like to keep `no_cache` the way it is for three reasons:
1) It removes any ambiguity about what `no_cache` means, keeping it simple that the cache is simply not used.

2) The only real use-cace I can think of for `no_cache` is when combined with
non-default `reader_args` values.

3) There could be a test which uses a large input file which is not used by any-other tests, allowing the source data to be de-allocated once the test is completed.

Alternately given reason 2, assuming all of the potential values in  `reader_args` are hashable, we could make `tuple(reader_args.items())` a part of the cache key.

mdemoret-nv: Works for me. Keep it as is

mdemoret-nv (approved):

PR #908, opened 2023-04-26
Fix early stopping code in dfencoder to use average loss of batches in validation set

### LINKED ISSUE #905: [BUG]: dfencoder early stopping code using last batch's loss instead of average loss of the epoch
### Version

23.03

_No response_

### Describe the bug.

The current implementation of early stopping code in dfencoder is using the net loss of the last batch of the validation set to evaluate when to early-stop. This approach can be problematic as the last batch of the validation set may contain a small number of data points, which may not be representative of the entire validation set. 
It is recommended to use the average loss of all batches in the validation set to evaluate the early stopping criteria, as this approach will be more objective and representative of the entire dataset. 
The bug should be fixed by updating the early stopping code to use the average loss of the batches in the validation set.

_No response_

### Relevant log output

_No response_

### Full env printout

_No response_

### Other/Misc.

_No response_

### Code of Conduct

hsin-c:
## Description
This PR fixes the early stopping code in dfencoder to use the average loss of all batches in the validation set, instead of just the net loss of the last batch. This change will ensure that the early stopping criteria are more representative of the entire validation set, rather than relying on a potentially unrepresentative subset of the data.

closes #905 

## Checklist


mdemoret-nv (changes requested):
Can you add a test that verifies if early stopping is working correctly?

hsin-c:
> Can you add a test that verifies if early stopping is working correctly?

Added testing for early stopping. (Providing coverage for the early-stopping code in general, not testing the change in this PR specifically as it's a logic update on the data science side).

mdemoret-nv (approved):
Thanks for adding that test

PR #912, opened 2023-04-28
Improve DFP period functionality to allow for better sampling and ignoring period

mdemoret-nv:
## Description
Currently, the DFP pipeline simulates time by breaking incoming messages up by a specific period and processing each period independently. This makes it impossible to process all of the incoming data at once for batch mode with a single trained model.

This PR adds a few things:

- If `DFPFileBatcherStage.period == None`, then all messages will be processed in a single batch, instead of per period
- Fixes how periods were handled to work with counts
   - Before, `"D"` would work as expected but `"5D"` would not. This was due to using `to_period`
- The `DFPFileBatcherStage.sampling_rate_s` property was deprecated in favor of a more general `sampling` property
   - This property can support different values
      - If its a string, the value is interpreted as a frequency. The first row for each frequency will be taken
      - If its a value between [0,1), its a fraction. A percentage of rows will be taken
      - If its >=1, its a count. A random count of rows will be taken

## Checklist

### DIFF
--- a/examples/digital_fingerprinting/production/morpheus/dfp/stages/dfp_file_batcher_stage.py
+++ b/examples/digital_fingerprinting/production/morpheus/dfp/stages/dfp_file_batcher_stage.py
@@ -78,7 +78,6 @@ class DFPFileBatcherStage(SinglePortStage):
         file_objs = []
 
         # Determine the date of the file, and apply the window filter if we have one
-        ts_and_files = []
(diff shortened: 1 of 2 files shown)


mdemoret-nv:
@drobison00 Do any of the DFP modules need to be updated to match?

drobison00 (approved):

PR #935, opened 2023-05-02
Fix race condition in CompareDataFrameStage

dagardner-nv:
## Description
* Fixes intermittent segfault in `CompareDataFrameStage` and `ValidationStage`

## Checklist

### DIFF
--- a/morpheus/stages/output/compare_dataframe_stage.py
+++ b/morpheus/stages/output/compare_dataframe_stage.py
@@ -107,7 +107,7 @@ class CompareDataFrameStage(InMemorySinkStage):
         if self._reset_index:
             combined_df.reset_index(inplace=True)
 
-        retults = compare_df.compare_df(self._compare_df,
+        results = compare_df.compare_df(self._compare_df,
                                         combined_df,
                                         include_columns=self._include_columns,
                                         exclude_columns=self._exclude_columns,
@@ -118,4 +118,4 @@ class CompareDataFrameStage(InMemorySinkStage):
         if clear:
             self.clear()
 
-        return retults
+        return results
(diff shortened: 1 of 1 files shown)


morpheus/stages/output/compare_dataframe_stage.py
```diff
@@ -107,13 +107,15 @@ def get_results(self, clear=True) -> dict:
         if self._reset_index:
             combined_df.reset_index(inplace=True)
 
+        retults = compare_df.compare_df(self._compare_df,
```
drobison00, line 110:
```suggestion
        results = compare_df.compare_df(self._compare_df,
```

morpheus/stages/output/compare_dataframe_stage.py
```diff
@@ -107,13 +107,15 @@ def get_results(self, clear=True) -> dict:
...
             self.clear()
 
-        return compare_df.compare_df(self._compare_df,
-                                     combined_df,
-                                     include_columns=self._include_columns,
-                                     exclude_columns=self._exclude_columns,
-                                     replace_idx=self._index_col,
-                                     abs_tol=self._abs_tol,
-                                     rel_tol=self._rel_tol)
+        return retults
```
drobison00, line 121:
```suggestion
        return results
```

drobison00 (approved):

cwharris (approved):
I wonder if `get_messages` should return a copy of the message array rather than an instance of it...

dagardner-nv:
> I wonder if `get_messages` should return a copy of the message array rather than an instance of it...

I considered that, but decided against it since:
1) It would come at a higher cost.
2) Calling `clear` should be self descriptive.

PR #942, opened 2023-05-08
Add support for fsspec.core.OpenFile instances to the MonitorStage

### LINKED ISSUE #941: [BUG]: Pipeline hangs when Monitor stages are enabled with unsupported types
### Version

23.07

Docker

### Describe the bug.
This happens when a monitor stage is inserted after one of the stages that emits an `fsspec` object instead of a message.

By default the pipeline runs with the log level set to `WARN` which omits the monitor stages from the pipeline, the inference pipeline completes in about 45 seconds, and training the generic user completes in about 20 seconds.

When the monitor stages are added via `--log_level INFO` or higher, the pipeline hangs. What appears to be happening is a race condition where the `MultiFileSource` stage emits file objects before the monitor stage starts.

```shell
python dfp_duo_pipeline.py --train_users none --start_time "2022-08-01" --input_file="../../../data/dfp/duo-training-data/*.json" --log_level INFO
```

### Relevant log output

_No response_

### Full env printout

_No response_

### Other/Misc.

_No response_

### Code of Conduct

dagardner-nv:
## Description
* Allows the `MonitorStage` to be placed immediately after stages such as `MultiFileSource` which emits `fsspec.core.OpenFiles` instead of message types.
* Adds more tests for `MonitorStage`, some of these test that when the stage does raise a `NotImplementedError` that the pipeline halts propperly which was ultimately fixed in MRC PRs https://github.com/nv-morpheus/MRC/pull/326 and https://github.com/nv-morpheus/MRC/pull/327

fixes #941

## Checklist


mdemoret-nv (changes requested):
I'm not opposed to these changes, but I'm concerned that they will hide bigger issues with unsupported types all defaulting to `1`. Can we figure out why this exception isnt causing the pipeline to shutdown correctly?

At the very least, I think a warning message should be generated like "The monitor stage does not recognize type XYZ. All messages will default to a count of 1. If this is unexpected, specify `determine_count_fn` in the MonitorStage constructor or update `MonitorController.auto_count_fn`"

tests/test_pipe_error.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env python
```
mdemoret-nv, line 1: Lets rename this to `test_error_pipe.py` to match other names

mdemoret-nv (approved):
Merge once comments are resolved

PR #946, opened 2023-05-10
Add COLLABORATOR to exclusion in label-external-issues.yml

jarmak-nv:
## Description
Sometimes `github.event.issue.authorAssociation` in the workflow will pick `COLLABORATOR` for our team members. This PR adds `COLLABORATOR` to the list of association Enums that we skip the triage label action for.

## Checklist

### DIFF
--- a/.github/workflows/label-external-issues.yml
+++ b/.github/workflows/label-external-issues.yml
@@ -27,7 +27,7 @@ jobs:
   Label-Issue:
     runs-on: ubuntu-latest
     # Only run if the issue author is not part of NV-Morpheus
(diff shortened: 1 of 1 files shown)


drobison00:
Why does it only do this sometimes?

jarmak-nv:
Excellent question, one I don't have an answer for. I think right now it just might be due to the [API issues GH is having at the moment](https://www.githubstatus.com/), causing certain endpoints to not respond and returning a different than normal value.

I've asked this question and not gotten a direct response. IMO it should return a list of all values the author qualifies for: https://docs.github.com/en/graphql/reference/enums#commentauthorassociation (all other authorAssociations draw from the CommentAuthorAssociation Enum).

mdemoret-nv (approved):

Issue #508, opened 2022-11-29
[DOC]: CLX/Morpheus Parity Comparison - Engineering


BartleyR:
### How would you describe the priority of this documentation request

Medium

### Describe the future/missing documentation

The current RAPIDS CLX library has functionality that could be integrated into Morpheus. As the first step towards this, we wish to understand the parity between CLX and Morpheus in terms of the APIs, capabilities, developer tools, and methods that exist in both.

### Where have you looked?

No such documentation exist in CLX or Morpheus

### Tasks
Generated from @efajardo-nv's comment below

```[tasklist]
- [x] clx.analytics.loda.Loda
- [x] clx.analytics.multiclass_sequence_classifier.MulticlassSequenceClassifier
- [x] clx.parsers.splunk_notable_parser.SplunkNotableParser
- [x] clx.parsers.windows_event_parser.WindowsEventParser
- [x] clx.parsers.zeek.parse_log_file
- [x] clx.analytics.dga_detector.DGADetector
- [x] clx.dns.dns_extractor
- [x] clx.ip
```
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I have searched the [open documentation issues](https://github.com/nv-morpheus/Morpheus/issues?q=is%3Aopen+is%3Aissue+label%3Adoc) and have found no duplicates for this bug report

efajardo-nv:
Possible CLX functionality to migrate:

LODA anomaly detection
- `clx.analytics.loda.Loda`
- Used in [Network Intrusion Detection in Netflow](https://github.com/rapidsai/clx/blob/branch-23.02/notebooks/ids_detection/IDS_using_LODA.ipynb)

Multi-class Sequence classifier
- `clx.analytics.multiclass_sequence_classifier.MulticlassSequenceClassifier`
- Can be used for various classifier use cases such as [Root Cause Analysis](https://github.com/nv-morpheus/Morpheus/tree/branch-23.01/examples/root_cause_analysis) which currently uses binary sequence classifier.

Splunk Notable Parser
- `clx.parsers.splunk_notable_parser.SplunkNotableParser`
- Used in [Analysis of Alert Notables from Splunk](https://github.com/rapidsai/clx/blob/branch-23.02/notebooks/alert_analysis/Alert_Analysis_with_CLX.ipynb)

Windows Event Log Parser
- `clx.parsers.windows_event_parser.WindowsEventParser`

Zeek Parser
[...]

PR #952, opened 2023-05-18
Multi-class Sequence Classifier

efajardo-nv:
## Description
- Migrate CLX multi-class sequence classifier. Update existing binary classifier to support more than two labels.
- Update existing phishing and root cause models to use models generated from updated sequence classifier.
- Remove duplicated Python files from `training-tuning-scripts`. Add single version of these files to `common` directory and symlinks to `common` for phishing and root cause training.
- Remove duplicated tokenizer hash files. Update training scripts to point to hash file in `morpheus/data`.
- Update phishing and root cause training/inference scripts and notebooks.

Task from issue #508

## Checklist
[x] I am familiar with the [Contributing Guidelines](https://github.com/nv-morpheus/Morpheus/blob/main/docs/source/developer_guide/contributing.md).
[x] New or existing tests cover these changes.
[x] The documentation is up to date with these changes.


models/training-tuning-scripts/common/dataloader.py
```diff
@@ -0,0 +1,52 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.
```
mdemoret-nv, line 1: Lets add a `__init__.py` to the `common` directory to make it a python module and then lets symbolically link the entire `common` folder into each subfolder that needs it instead of linking the individual files that are needed.

mdemoret-nv (changes requested):

mdemoret-nv (approved):

Issue #934, opened 2023-05-01
[BUG]: tests/test_dfp.py intermittently Segmentation faults


dagardner-nv:
### Version

23.07

### Which installation method(s) does this occur on?

Source

### Describe the bug.

```
============================================================================ test session starts ============================================================================
platform linux -- Python 3.10.10, pytest-7.3.1, pluggy-1.0.0
benchmark: 4.0.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /home/dagardner/work/morpheus
configfile: pyproject.toml
testpaths: tests
plugins: Faker-12.3.0, cov-4.0.0, anyio-3.6.2, benchmark-4.0.0, kafka-0.5.1
collected 1105 items                                                                                                                                                        

tests/test_abp.py ....                                                                                                                                                [  0%]
tests/test_abp_kafka.py ss                                                                                                                                            [  0%]
tests/test_add_classifications_stage.py ...                                                                                                                           [  0%]
tests/test_add_classifications_stage_pipe.py ....                                                                                                                     [  1%]
tests/test_add_scores_stage.py ...                                                                                                                                    [  1%]
tests/test_add_scores_stage_pipe.py ..........................................                                                                                        [  5%]
tests/test_appshield_source_stage.py ................                                                                                                                 [  7%]
tests/test_cli.py ..............                                                                                                                                      [  8%]
tests/test_column_info.py .........                                                                                                                                   [  9%]
tests/test_concat_df.py ...                                                                                                                                           [  9%]
tests/test_config.py .........                                                                                                                                        [ 10%]
tests/test_conftest.py ........................................................                                                                                       [ 15%]
tests/test_deserialize_stage_pipe.py ..........                                                                                                                       [ 16%]
tests/test_dfp.py Segmentation fault
```

### Minimum reproducible example

```shell
pytest --run_slow
```

### Code of Conduct

- [X] I agree to follow Morpheus' Code of Conduct
- [X] I have searched the [open bugs](https://github.com/nv-morpheus/Morpheus/issues?q=is%3Aopen+is%3Aissue+label%3Abug) and have found no duplicates for this bug report

dagardner-nv:
[...]

PR #955, opened 2023-05-19
Remove/update tests which mock MRC builder objects

dagardner-nv:
## Description
* Remove tests weren't testing anything not covered in an end-to-end test, and had the possibility of triggering issue #934
* If the test was checking legitimate functionality the test was updated such that MRC internals weren't left half mocked.
* Ignore `import-error` and `no-member` errors, both of these are erroneous errors on pylints part.
* Fix what appears to be a variable name change in MRC Builder of `in_port_streams` to `to in_ports_streams`

## Checklist


pyproject.toml
```diff
@@ -434,16 +434,19 @@ disable = [
   "bad-inline-option",
   "deprecated-pragma",
   "file-ignored",
+  "import-error", # pylint gets confused by tests for our examples
```
mdemoret-nv, line 437: Dont love this. Would prefer to use `--generated_members` if needed

pyproject.toml
```diff
@@ -434,16 +434,19 @@ disable = [
...
   "import-outside-toplevel", # Allow lazy imports inside of methods
   "locally-disabled",
   "missing-class-docstring",
   "missing-function-docstring",
   "missing-module-docstring",
+  "no-member", # pylint isn't aware of some of our attrs
```
mdemoret-nv, line 443: Lets make an issue to address this with changing the check and build stage order in CI

mdemoret-nv (approved):

pyproject.toml
dagardner-nv: When I disable this warning, it fails on:
```py
def test_constructor(config: Config):
    from dfp.stages.write_to_s3_stage import WriteToS3Stage
```
```
tests/examples/digital_fingerprinting/test_write_to_s3_stage.py:23:4: E0401: Unable to import 'dfp.stages.write_to_s3_stage' (import-error)
```

I can work-around this issue by running:
```sh
PYTHONPATH="examples/digital_fingerprinting/production/morpheus" ./ci/scripts/python_checks.sh
```

We would need to do this for the ransomeware example and the `models/training-tuning-scripts` dir. The potential problem would occur if any of these have a name conflict.

PR #977, opened 2023-06-08
Adding HTTP sources & sinks

### LINKED ISSUE #864: [FEA]: Create REST Source/Sink Stages

New Feature

High

To make a Morpheus pipeline that can receive messages from a client requires making a custom stage. While the implementation of a REST source and sink will vary significantly, we can provide a base implementation which is extendable for users needs.

This new feature should include:

- A new source object `RestSourceStage`
   - This new stage will be capable of receiving `POST` messages on a specific endpoint and converting the body into a DataFrame
   - The IP address, port, and endpoint should be configurable
   - [Optional] Configuration options for authentication and SSL should be provided
   - The body of each `POST` message will be fed directly into the `cudf.read_json` method to convert into a DataFrame
- A new sink object `RestSinkStage`
   - This new sink will be capable of converting a DataFrame into a `POST` message and sent to a specific endpoint
   - The IP address, port, and endpoint should be configurable
   - There will be 2 modes this can run in:
      - DataFrame mode: The entire DataFrame will be converted to JSON and will be used as the body of the `POST` message
      - Row-by-row mode: A template will be supplied by the user for the endpoint, headers, and body of the message with keyword options. Each row will use the template to generate the endpoint, headers, and body of the message with the columns being used for the keywords
- An example will be added to the `examples` directory to show how a custom REST source/sink could be used which performs more complex logic (i.e. authentication or session management)
- Tests for the new stages
- Documentation for the new stages

_No response_

### Additional context

_No response_

### Code of Conduct

dagardner-nv:
## Description
* Adds four new stages (#864):
  1. HttpServerSourceStage - Spins up a rest server to listen for incoming POSTS
  2. HttpClientSourceStage - Source that polls a remote HTTP server for incoming data
  3. HttpClientSinkStage - Sink that POSTS data to a remote HTTP server
  4. HttpServerSinkStage - Sink that statrs a rest server listening for GET requests
* Fix `TensorIndex` usage for the `stop_after` feature of the KafkaSourceStage

fixes #864

Currently `HttpServerSourceStage` is the only one of the four with a C++ impl. The rest server itself is written in C++ with Python bindings and is used by both C++ & Python stages.

## Checklist

### DIFF
--- a/tests/test_http_client_source_stage_pipe.py
+++ b/tests/test_http_client_source_stage_pipe.py
@@ -27,7 +27,8 @@ from morpheus.stages.output.compare_dataframe_stage import CompareDataFrameStage
 @pytest.mark.parametrize("lines", [False, True])
 def test_http_client_source_stage_pipe(config: Config, dataset: DatasetManager, mock_rest_server: str, lines: bool):
     """
-    The data is read from the mock REST server and compared against the expected data.
+    Test the HttpClientSourceStage against a mock REST server which will return JSON data which can be deserialized
+    into a DataFrame.
     """
     if lines:
         endpoint = "data-lines"


morpheus/utils/http_utils.py
```diff
@@ -0,0 +1,115 @@
...
+logger = logging.getLogger(__name__)
+
+
+class MimeTypes(Enum):
+    """Not a complete list of mime types, just the ones we use."""
+    TEXT = "text/plain"
+    JSON = "application/json"
+
+
+def request(
```
mdemoret-nv, line 32: `request_with_retry` might be a better name here to differentiate it.

morpheus/utils/http_utils.py
```diff
@@ -0,0 +1,115 @@
...
+
+            logger.error("Error occurred performing %s request to %s: %s",
+                         request_kwargs['method'],
+                         request_kwargs['url'],
+                         e)
+            logger.debug("Sleeping for %s seconds before retrying request again", sleep_time)
+            time.sleep(sleep_time_)
+
+
+def verify_url(url: str) -> str:
```
mdemoret-nv, line 100: "verify" implies its going to return a True/False if it passed verification. `prepare_url` or `url_ensure_protocol` would be better names.

morpheus/stages/input/rest_source_stage.py
```diff
@@ -0,0 +1,198 @@
...
+                logger.error("Queue closed unexpectedly, shutting down")
+                processing = False
+
+            if df is not None:
+                yield MessageMeta(df)
+
+        logger.debug("Stopping REST server ...")
+        rest_server.stop()
+        logger.debug("REST server stopped")
+        self._queue.close()
```
mdemoret-nv, line 177: Wrap this in a try...finally block to ensure the server gets closed.

mdemoret-nv: Or an even better option would be to have `RestServer` be a context manager

morpheus/utils/http_utils.py
mdemoret-nv, line 32: If the `RestServer` is entirely implemented in C++, can we have a similar client that works the same way? Then we wouldnt even need to use `requests` at all.

morpheus/_lib/src/utilities/rest_server.cpp
```diff
@@ -0,0 +1,436 @@
...
+{
+    CHECK(!m_is_running) << "RestServer is already running";
+
+    try
+    {
+        DLOG(INFO) << "Starting RestServer on " << m_bind_address << ":" << m_port << " with " << m_num_threads
+                   << " threads";
+        m_is_running = true;
+        m_listener_threads.reserve(m_num_threads);
+        m_listener_threads.emplace_back(std::thread(&RestServer::start_listener, this));
```
mdemoret-nv, line 329: In general, we want to use MRC to manage all threads/fibers so they can be managed by the runtime. But this should be ok. Going to look and see if there is any way to better link the boost thread creation with MRC.

Main thing I am worried about here is MRC starts every thread with some initializer code to allow for debugging python. Without it, breakpoints and things wont work. Since this does a callback into python, you wouldnt hit any breakpoints.

morpheus/_lib/src/utilities/rest_server.cpp
```diff
@@ -0,0 +1,436 @@
...
+    auto ioc     = m_io_context;  // ensure each thread gets its own copy including this one
+
+    std::make_shared<Listener>(
+        ioc, m_payload_parse_fn, m_bind_address, m_port, m_endpoint, m_method, m_max_payload_size, m_request_timeout)
+        ->run();
+
+    for (auto i = 1; i < m_num_threads; ++i)
+    {
+        m_listener_threads.emplace_back([ioc]() { ioc->run(); });
+    }
```
mdemoret-nv, line 314: If `num_threads==1`, will any listeners be created? Does it still function with 1 thread?

mdemoret-nv (commented):
Overall, the design looks great. Some small concerns about how threads are created but everything should be ok.

morpheus/_lib/src/objects/fiber_queue.cpp
```diff
@@ -111,18 +111,10 @@ void FiberQueueInterfaceProxy::put(morpheus::FiberQueue& self, pybind11::object
...
     case boost::fibers::channel_op_status::success:
         return;
-    case boost::fibers::channel_op_status::empty: {
-        // Raise queue.Empty
-        pybind11::object exc_class = pybind11::module_::import("queue").attr("Empty");
-
-        PyErr_SetNone(exc_class.ptr());
-
-        throw pybind11::error_already_set();
-    }
```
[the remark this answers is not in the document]
dagardner-nv: The `FiberQueue::put` will never return a channel status of empty. The method calls one of three methods `try_push`, `push_wait_for` and `push` and according to the documentation they should never return an empty.

Maybe this should be changed to a `DCHECK`?

morpheus/stages/input/rest_source_stage.py
```diff
@@ -0,0 +1,198 @@
...
+        If False, the REST server will expect each request to be a JSON array of objects. If True, the REST server will
+        expect each request to be a JSON object per line.
+    """
+
+    def __init__(self,
+                 config: Config,
+                 bind_address: str = "127.0.0.1",
+                 port: int = 8080,
+                 endpoint: str = "/message",
+                 method: str = "POST",
```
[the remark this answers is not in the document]
dagardner-nv: Agreed.

morpheus/_lib/src/utilities/rest_server.cpp
dagardner-nv: The thread this method is executed in is the first worker and executes `ioc->run()` just after this loop, and the current thread is already in the 0th position in `m_listener_threads` which is asserted at the top of this method.

morpheus/utils/http_utils.py
dagardner-nv: Yes, we could. 

My initial thought was that a C++ impl for a client wouldn't improve performance too much over that of a Python client. Though I think you pointed out it could still offer some other benefits in other parts of the code base.

morpheus/stages/input/rest_source_stage.py
dagardner-nv: Done, but now I'm wondering if we should add a context manager to `FiberQueue`

cwharris:
I was thinking we could consider the following names.
```
RestSourceStage - HttpHookSourceStage
RestClientSourceStage - HttpRequestSourceStage
WriteToRestStage - HttpRequestSinkStage
RestServerSinkStage - HttpHookSinkStage
```

I know `HttpRequestSinkStage` differs from "WriteTo*", but it's very explicit about what it's doing, and creates good uniformity between the different HTTP-related stages.

tests/test_rest_client_source_stage_pipe.py
```diff
@@ -0,0 +1,44 @@
...
+from morpheus.stages.input.rest_client_source_stage import RestClientSourceStage
+from morpheus.stages.output.compare_dataframe_stage import CompareDataFrameStage
+from utils import assert_results
+from utils.dataset_manager import DatasetManager
+
+
+@pytest.mark.slow
+@pytest.mark.use_cudf
+@pytest.mark.parametrize("lines", [False, True])
+def test_rest_client_source_stage_pipe(config: Config, dataset: DatasetManager, mock_rest_server: str, lines: bool):
```
mdemoret-nv, line 28: Add a comment where it gets it's data from

mdemoret-nv (approved):
Lets make an issue to add C++ implementations using the REST client loader

dagardner-nv:
> Lets make an issue to add C++ implementations using the REST client loader

Done #1150

PR #986, opened 2023-06-14
Implement rest data loader

yuchenz427:
## Description
Implement RESTDataLoader.

## Checklist


morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,238 @@
...
+            {
+                break;
+            }
+            std::this_thread::sleep_for(std::chrono::milliseconds(interval_milliseconds));
+            interval_milliseconds *= 2;
+        }
+
+    } catch (std::exception const& e)
+    {
+        // TODO: need VLOG here?
```
drobison00, line 141: Yes, please. Good to avoid printing to any standard streams directly and let the logger handle it.

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,238 @@
...
+        target = "/";
+    }
+    std::string query = result.attr("query").is_none() ? "" : result.attr("query").cast<std::string>();
+    // if params exists, overrides query part included in endpoint
+    if (!params.empty())
+    {
+        query = params;
+    }
+    py::gil_scoped_release gil_release;
+
```
drobison00, line 86: The same thing can be accomplished with a local nested scope, which is a bit cleaner/easier to read, and avoids having to explicitly release the gil (it will automatically go out of scope).

```suggestion
    std::string ep{endpoint};
    std::string query;    

    {
        py::gil_scoped_acquire gil;

        py::module_ urllib = py::module::import("urllib.parse");
        py::object ep_pystr = py::str(ep);

        // Following the syntax specifications in RFC 1808, urlparse recognizes a netloc only if it is properly introduced
        // by ‘//’. Otherwise the input is presumed to be a relative URL and thus to start with a path component. 
        // ref: https://docs.python.org/3/library/urllib.parse.html
        // Workaround here is to prepend "//" if endpoint does not already include one
        if (ep_pystr.attr("find")("//").cast<int>() == -1)
        {
            ep = "//" + ep;
        }

        py::object result  = urllib.attr("urlparse")(ep);
        std::string host   = result.attr("hostname").is_none() ? "" : result.attr("hostname").cast<std::string>();
        std::string target = result.attr("path").is_none() ? "" : result.attr("path").cast<std::string>();
        if (target.empty())
        {
            target = "/";
        }

        // if params exists, overrides query part included in endpoint
        std::string query = result.attr("query").is_none() ? "" : result.attr("query").cast<std::string>();
        if (!params.empty())
        {
            query = params;
        }
    }
```

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,238 @@
...
+    // Aggregate dataframes for each file
+    py::gil_scoped_acquire gil;
+    py::module_ mod_cudf;
+    py::object dataframe = py::none();
+
+    auto& cache_handle = mrc::pymrc::PythonObjectCache::get_handle();
+    mod_cudf           = cache_handle.get_module("cudf");
+    py::gil_scoped_release release;
+
+    // TODO(Devin) : error checking + improve robustness
```
drobison00, line 191:
```suggestion
```

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,238 @@
...
+    std::string strategy = task.value("strategy", "aggregate");
+    if (strategy != "aggregate")
+    {
+        throw std::runtime_error("Only 'aggregate' strategy is currently supported");
+    }
+
+    auto conf     = this->config();
+    int max_retry = conf.value("max_retry", 3);
+    auto queries  = task["queries"];
+    // TODO(Devin) : Migrate this to use the cudf::io interface
```
drobison00, line 206:
```suggestion
```

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,238 @@
...
+    for (auto& query : queries)
+    {
+        std::string method = query.value("method", "GET");
+        std::transform(method.begin(), method.end(), method.begin(), ::toupper);
+        std::string endpoint = query.value("endpoint", "");
+        if (endpoint.empty())
```
drobison00, line 212: Should this be an error?

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,238 @@
...
+}
+
 std::shared_ptr<ControlMessage> RESTDataLoader::load(std::shared_ptr<ControlMessage> message, nlohmann::json task)
 {
     VLOG(30) << "Called RESTDataLoader::load()";
 
-    // TODO(Devin): Implement this
-    throw std::runtime_error("RESTDataLoader::load() not implemented yet");
+    // Aggregate dataframes for each file
+    py::gil_scoped_acquire gil;
```
drobison00, line 183: Having to do something like 'gil' and 'gil_2' is a bit clunky here. There are a couple possible workarounds. One is to adapt this and the piece at the bottom to use locally nested scope blocks, like the example I'd given above. You'll need to be a bit more careful though, since you want to fill in python objects.

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,238 @@
...
+        std::transform(method.begin(), method.end(), method.begin(), ::toupper);
+        std::string endpoint = query.value("endpoint", "");
+        if (endpoint.empty())
+        {
+            continue;
+        }
+        std::string content_type = query.value("content_type", "");
+        std::string body         = query.value("body", "");
+        nlohmann::json x_headers = query.value("x-headers", nlohmann::json());
+        std::unordered_map<std::string, std::string> x_headers_map;
```
drobison00, line 219: Since there is a decent amount of work going on here, it will probably be worth it to make this section more declarative as well.

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,238 @@
...
+                param_str.pop_back();
+                http::response<http::dynamic_body> response;
+                get_data_from_endpoint(
+                    response, method, endpoint, param_str, content_type, body, x_headers_map, max_retry);
+                create_dataframe_from_http_response(response, dataframe, mod_cudf, strategy);
+            }
+        }
+    }
+
+    py::gil_scoped_acquire gil_2;
```
drobison00, line 250: See comment above.

tests/modules/test_morpheus_modules.py
```diff
@@ -345,10 +344,99 @@ def _on_next(control_msg):
...
+            config = {
+                "tasks": [{
+                    "type": "load",
+                    "properties": {
+                    "loader_id": "rest", 
+                    "strategy": "aggregate", 
+                    "queries": [{
+                    "method": "GET",
+                    "endpoint": "0.0.0.0/?param1=false&param2=true",
+                    "params": [
```
drobison00, line 365: We'll want to document these parameters, see:

docs/source/loaders

tests/modules/test_morpheus_modules.py
```diff
@@ -345,10 +344,99 @@ def _on_next(control_msg):
     for f in files:
         os.remove(f[0])
 
+def test_rest_loader_module():
```
drobison00, line 347: Lets expand this out a bit; you're testing a few different things in this one unit test. I'd like to see a few different scenarios spread across multiple tests.

tests/modules/test_morpheus_modules.py
```diff
@@ -245,7 +245,6 @@ def _on_next(control_msg):
```
drobison00, line 245: We'll need to sync on adding additional tests that use the camouflage service.  See David's upcoming PR: https://github.com/nv-morpheus/Morpheus/pull/977/files#diff-13ada51897e5cf3d7ca958ffbebce048efecf04fe47e44de8f97a045b650cce3

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,238 @@
...
+        query = params;
+    }
+    py::gil_scoped_release gil_release;
+
+    net::io_context ioc;
+    tcp::resolver resolver(ioc);
+    beast::tcp_stream stream(ioc);
+    try
+    {
+        auto const endpoint_results = resolver.resolve(host, PORT);
```
drobison00, line 92: I'd like to see this made a bit more declarative, it improves readability and makes updating things easier in the future.

For example:

```c++
auto verb = get_http_verb(method);
auto request = construct_request(verb, target, query, count);

init_request(request);

try_get_data(request, response, max_retry, interval, ...);
```

drobison00 (changes requested):
Looks good overall, I've got a few stylistic requests, docs updates, and then we'll need to sync with David about using his camouflage services to test against.

morpheus/_lib/src/io/loaders/rest.cpp
yuchenz427: I agree. The issue here is if locally nested scope block is adapted to the first _gil_acquire_, both `mod_cudf` and `dataframe` will turn into local variable, which get out of scope at the end of the block. The workaround I could think of is to keep the explicit `gil_release` for the first `gil_acquire`, then use locally nested scope to the second `gil_acquire` to avoid something like `gil_2`

morpheus/_lib/src/io/loaders/rest.cpp
yuchenz427: Agree. `Throw std::runtime_error` here

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,295 @@
...
+        if (status != 503 && status != 504)
+        {
+            break;
+        }
+        std::this_thread::sleep_for(std::chrono::milliseconds(interval_milliseconds));
+        interval_milliseconds *= 2;
+    }
+}
+
+void get_data_from_endpoint(http::response<http::dynamic_body>& response,
```
drobison00, line 147: It would be beneficial to split this function into smaller, more focused functions.

A helpful guideline is that when a function is performing operations that are not directly related to its stated purpose, these operations should be moved into separate functions. This enhances readability and maintainability.

In the case of the 'get_data_from_endpoint' function, it is not just retrieving data, but also parsing and formatting URLs. Therefore, it may be more effective to restructure the function as follows:

    get_data_from_endpoint
    parse_and_format_URL
    get_data

This approach will ensure that each function is focused on a single responsibility, making the code easier to understand and debug.

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,295 @@
...
+    net::io_context ioc;
+    tcp::resolver resolver(ioc);
+    beast::tcp_stream stream(ioc);
+    try
+    {
+        auto const endpoint_results = resolver.resolve(host, PORT);
+        auto verb                   = get_http_verb(method);
+        auto request = construct_request(verb, target, query, host, method, content_type, body, x_headers);
+
+        try_get_data(stream, endpoint_results, request, response, max_retry, 1000);
```
drobison00, line 200: I am not extremely familiar with beast, so I don't know if it does url-encoding by default; but, since you're using urllib, it may be worth thinking a bit about these issues:

https://docs.python.org/3/library/urllib.parse.html#url-parsing-security
https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,295 @@
...
+
+    } catch (std::exception const& e)
+    {
+        VLOG(30) << "Error when calling RESTDataLoader::load(): " << e.what();
```
drobison00, line 204: Fail gracefully if processes_failures_as_errors is false.

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,295 @@
...
+                                         py::module_& mod_cudf,
+                                         const std::string& strategy)
+{
+    std::string df_json_str = beast::buffers_to_string(response.body().data());
+    boost::algorithm::trim(df_json_str);
+
+    // When calling cudf.read_json() with engine='cudf', it expects an array object as input.
+    // The workaround here is to add square brackets if the original data is not represented as an array.
+    // Submitted an issue to cudf: https://github.com/rapidsai/cudf/issues/13527.
+    if (!(df_json_str.front() == '[' && df_json_str.back() == ']'))
```
drobison00, line 219: Does this work for patterns like this?: `\s*[.*]\s*`

We likely want to strip any leading whitespace, carriage returns, etc..

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,295 @@
...
+
+namespace beast = boost::beast;
+namespace http  = beast::http;
+namespace net   = boost::asio;
+using tcp       = net::ip::tcp;
+
+#define PORT "8081"
+
 namespace morpheus {
 RESTDataLoader::RESTDataLoader(nlohmann::json config) : Loader(config) {}
```
drobison00, line 48: As a general comment, lets be a bit more explicit about error handling in each section of the code here, and I'll make an issue to do the same for other existing loaders. In many cases, _any_ calls to python code could fail, so lets be sure all of them are wrapped in a try/catch block and exceptions are processed as expected when `processes_failures_as_errors` is false, otherwise we throw.

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,295 @@
...
+                          std::string& content_type,
+                          std::string& body,
+                          std::unordered_map<std::string, std::string> x_headers,
+                          nlohmann::json& params)
+{
+    method = query.value("method", "GET");
+    std::transform(method.begin(), method.end(), method.begin(), ::toupper);
+
+    endpoint = query.value("endpoint", "");
+    if (endpoint.empty())
```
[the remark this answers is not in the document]
yuchenz427: Implemented in `process_failures()`

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,378 @@
...
+
 std::shared_ptr<ControlMessage> RESTDataLoader::load(std::shared_ptr<ControlMessage> message, nlohmann::json task)
 {
     VLOG(30) << "Called RESTDataLoader::load()";
 
-    // TODO(Devin): Implement this
-    throw std::runtime_error("RESTDataLoader::load() not implemented yet");
+    // If set to false, any exception thrown during the task is caught and the related fields in ControlMessage are set
+    // to indicate the reason of that failure; Otherwise, the exception is thrown
+    bool processes_failures_as_errors = false;
```
drobison00, line 336: Could you actually extract this capture process to the base DataLoader implementation in `data_loader.cpp`, then this can form the basis for gracefully handling all loader failures instead of just this one.

std::shared_ptr<ControlMessage> DataLoader::load(std::shared_ptr<ControlMessage> control_message)

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,378 @@
...
+            py::gil_scoped_acquire gil;
+            message->payload(MessageMeta::create_from_python(std::move(dataframe)));
+        }  // release GIL
+    } catch (std::runtime_error& e)
+    {
+        process_failures(e.what(), message, processes_failures_as_errors);
+    }
+    return message;
 }
 }  // namespace morpheus
```
drobison00, line 394:
```suggestion
}  // namespace morpheus

```

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,378 @@
```
drobison00, line 17: One high level comment:

If we have functions that are only ever intended to be called within this `rest.cpp` compilation unit, please put them inside an anonymous namespace. Otherwise, while they're not part of the public API, their symbols are still accessible, and it is also possible to violate ODR within the project in a non-obvious way.

See my comment on the `convert_url_query_to_dict` function.

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,378 @@
...
+            {
+                throw std::runtime_error("'REST Loader' failed to parse URL query: " + raw_query.cast<std::string>());
+            }
+            params_dict[split_by_eq[0]] = split_by_eq[1];
+        }
+    }
+    return params_dict;
+}
+
+void parse_endpoint_to_url(
```
drobison00, line 178: Update based on public vs private handling discussion.

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,378 @@
...
+        for (auto& param : params.items())
+        {
+            query_pydict[py::str(param.key())] = py::str(param.value());
+        }
+    }
+    // Encode URL query
+    query = py::len(query_pydict) == 0 ? "" : "?" + urllib.attr("urlencode")(query_pydict).cast<std::string>();
+}
+
+void get_response_from_endpoint(http::response<http::dynamic_body>& response,
```
drobison00, line 222: Update based on public vs private handling discussion.

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,378 @@
...
+    std::string host;
+    std::string target;
+    std::string query;
+    http::request<http::string_body> request;
+    parse_endpoint_to_url(endpoint, params, host, target, query);
+    construct_request(request, method, host, target, query, content_type, body, x_headers);
+    get_response_with_retry(request, response, max_retry, retry_interval_milliseconds);
+}
+
+void create_dataframe_from_response(py::object& dataframe,
```
drobison00, line 241: Update based on public vs private handling discussion.

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,378 @@
...
+        {
+            py::list args;
+            args.attr("append")(dataframe);
+            args.attr("append")(current_df);
+            dataframe = mod_cudf.attr("concat")(args);
+        }
+    }  // release GIL
+}
+
+void create_dataframe_from_query(py::object& dataframe,
```
drobison00, line 276: Update based on public vs private handling discussion.

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,378 @@
...
+        request.body() = body;
+        request.prepare_payload();
+    }
+    for (auto& x_header : x_headers.items())
+    {
+        request.insert(x_header.key(), x_header.value());
+    }
+}
+
+void get_response_with_retry(http::request<http::string_body>& request,
```
drobison00, line 122: Update based on public vs private handling discussion.

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,378 @@
...
+        verb = http::verb::post;
+    }
+    else
+    {
+        throw std::runtime_error("'REST Loader' receives method not supported: " + method);
+    }
+    return verb;
+}
+
+void construct_request(http::request<http::string_body>& request,
```
drobison00, line 95: Update based on public vs private handling discussion.

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,378 @@
...
+        throw std::runtime_error("'REST loader' receives query with empty endpoint");
+    }
+
+    content_type = query.value("content_type", "");
+    body         = query.value("body", "");
+    params       = query.value("params", nlohmann::json());
+    x_headers    = query.value("x-headers", nlohmann::json());
+}
+
+http::verb get_http_verb(const std::string& method)
```
drobison00, line 77: Update based on public vs private handling discussion.

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,378 @@
...
+namespace http  = beast::http;
+namespace net   = boost::asio;
+using tcp       = net::ip::tcp;
+
+#define PORT "8081"
+
 namespace morpheus {
 RESTDataLoader::RESTDataLoader(nlohmann::json config) : Loader(config) {}
 
+void extract_query_fields(nlohmann::json& query,
```
drobison00, line 54: Update based on public vs private handling discussion.

drobison00 (changes requested):
Some quick structural changes, a couple issues with magic numbers, and some suggestions.

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,378 @@
...
+        http::read(stream, buffer, response);
+        beast::error_code ec;
+        stream.socket().shutdown(tcp::socket::shutdown_both, ec);
+        if (ec && ec != beast::errc::not_connected)
+        {
+            throw beast::system_error{ec};
+        }
+        int status = response.result_int();
+        // Retry only if status is 503 Service Unavailable / 504 Gateway Timeout
+        if (status != 503 && status != 504)
```
bsuryadevara, line 146: Using named constants for HTTP status codes `503` and `504` would be a good for reusability.

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,378 @@
...
+    // If not present in URL:
+    //  - host: None
+    //  - path/query: Empty Python String
+    host = result.attr("hostname").is_none() ? "" : result.attr("hostname").cast<std::string>();
+
+    // Encode URL target
+    target = urllib.attr("quote")(result.attr("path")).cast<std::string>();
+    target = target.empty() ? "/" : target;
+
+    py::str raw_query = result.attr("query");
```
bsuryadevara, line 204: Could you construct `host`, `target` and `raw_query` only once for multiple params set?

bsuryadevara (changes requested):
I've added a couple of comments. Nice work!

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,378 @@
...
+                       const std::string& target,
+                       const std::string& query,
+                       const std::string& content_type,
+                       const std::string& body,
+                       const nlohmann::json& x_headers)
+{
+    auto verb = get_http_verb(method);
+    request.method(verb);
+    request.target(target + query);
+    request.version(11);
```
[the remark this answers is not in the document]
yuchenz427: Decided to support http 1.1 & 1.0 here - relevant documentation is added here.

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,378 @@
...
+            break;
+        }
+        std::this_thread::sleep_for(std::chrono::milliseconds(retry_interval_milliseconds));
+        retry_interval_milliseconds *= 2;
+        max_retry--;
+    }
+}
+
+py::dict convert_url_query_to_dict(const py::str& raw_query)
+{
```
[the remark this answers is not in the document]
yuchenz427: Has made it private.

morpheus/_lib/src/io/data_loader.cpp
```diff
@@ -25,6 +25,21 @@
...
+                      std::shared_ptr<morpheus::ControlMessage> message,
+                      bool processes_failures_as_errors)
+{
+    if (processes_failures_as_errors)
+    {
+        throw std::runtime_error(error_msg);
+    }
+    message->set_metadata("cm_failed", "true");
+    message->set_metadata("cm_failed_reason", error_msg);
+    std::cout << "<<<<<<<<<<<<<<<<<< cm_failed_reason: " << error_msg << std::endl;
```
drobison00:
```suggestion

```

morpheus/_lib/src/io/data_loader.cpp
```diff
@@ -56,8 +80,13 @@ std::shared_ptr<ControlMessage> DataLoader::load(std::shared_ptr<ControlMessage>
         {
             VLOG(5) << "Loading data using loader: " << loader_id
                     << " for message: " << control_message->config().dump() << std::endl;
-
-            loader->second->load(control_message, task);
+            try
+            {
+                loader->second->load(control_message, task);
+            } catch (std::runtime_error& e)
```
drobison00: Catch all errors, not just std::runtime_error

morpheus/_lib/src/io/data_loader.cpp
```diff
@@ -56,8 +80,13 @@ std::shared_ptr<ControlMessage> DataLoader::load(std::shared_ptr<ControlMessage>
...
                     << " for message: " << control_message->config().dump() << std::endl;
-
-            loader->second->load(control_message, task);
+            try
+            {
+                loader->second->load(control_message, task);
+            } catch (std::exception& e)
+            {   
+                process_failures(e.what(), control_message, processes_failures_as_errors);
+            }
```
drobison00, line 89: Add an ellipsis catch (...), after std::exception, for unknown or non-standard errors.

morpheus/_lib/src/io/loaders/rest.cpp
```diff
@@ -17,21 +17,416 @@
...
+    //     https://www.boost.org/doc/libs/1_82_0/libs/beast/doc/html/beast/design_choices/faq.html)
+    //   - The default value of http_version is 1.1 if not specified within REST loader's config, since Beast sets
+    //     to 1.1 by default.
+    //   - Beast provides http client examples compatible with HTTP/1.0 (see
+    //     https://www.boost.org/doc/libs/1_82_0/libs/beast/example/http/client/sync/http_client_sync.cpp). REST loader
+    //     implements the http client in the same way as Beast's example, providing support to HTTP/1.0 if explicitly
+    //     specified in the config.
+    if (http_version != "1.1" && http_version != "1.0")
+    {
+        throw std::runtime_error("'REST Loader' receives http version not supported: " + http_version);
```
drobison00, line 121:
```suggestion
        throw std::runtime_error("'REST Loader' received http version not supported: " + http_version);
```

morpheus/_lib/src/io/data_loader.cpp
```diff
@@ -86,6 +86,8 @@ std::shared_ptr<ControlMessage> DataLoader::load(std::shared_ptr<ControlMessage>
             } catch (std::exception& e)
             {   
                 process_failures(e.what(), control_message, processes_failures_as_errors);
+            } catch (...) {
+                VLOG(5) << "DataLoader::load() caught unknown exception" << std::endl;
```
drobison00, line 90:
```suggestion
                process_failures("Unknown error", control_message, processes_failures_as_errors);
```

drobison00 (approved):

bsuryadevara (approved):

PR #1003, opened 2023-06-27
Fix pybind11 link errors

cwharris:
This explicitly sets `-Wl,--gc-sections`, which is automatically added by conda compilers. It is required to compile code containing pybind11 headers that has been optimized, as otherwise the linker will complain about an undefined hidden symbol.


morpheus/_lib/cmake/libmorpheus.cmake
```diff
@@ -82,12 +82,14 @@ target_link_libraries(morpheus
...
     ${cudf_helpers_target}
     matx::matx
+    
   PUBLIC
     cudf::cudf
     CUDA::nvToolsExt
     mrc::pymrc
     RDKAFKA::RDKAFKA
     TritonClient::httpclient_static
+    -Wl,--gc-sections # required to link code containing pybind11 headers
```
mdemoret-nv, line 92: You should use
```
target_link_options(morpheus PUBLIC "-Wl,--gc-sections")
```
Like we are already doing on line 141

mdemoret-nv (changes requested):

mdemoret-nv (approved):

PR #1024, opened 2023-07-07
Remove conda's base pip (CVE)

pdmack:
## Description
Marginally improve the CVE posture of the released image by removing conda's base pip exe.

## Checklist


docker/Dockerfile
```diff
@@ -116,6 +116,8 @@ RUN --mount=type=cache,id=conda_pkgs,target=/opt/conda/pkgs,sharing=locked \
     conda config --add pkgs_dirs /opt/conda/pkgs &&\
     # Install mamba, boa and git here. Conda build breaks with other git installs
     /opt/conda/bin/mamba install -y -n base -c conda-forge "boa" "git >=2.35.3" "python=${PYTHON_VER}" "tini=0.19" &&\
+    # CVE-2018-20225 for the base pip, not the env one
+    python -m pip uninstall -y pip && \
```
mdemoret-nv, line 120: Would it be better to use conda to remove pip here? Not sure if it makes a difference but using pip to remove itself seems weird to me.

dagardner-nv, line 120: Can you move this to the `runtime` stage, which I believe is the final stage for the images that we upload to NGC

mdemoret-nv, line 120: This seems early to me. Dont we only need to remove pip in the release container? This will remove pip for our development container which would be a pain. Somewhere around line 254 seems like a better place.

dagardner-nv (commented):
My big concern here is that some of the examples require additional dependencies such as `examples/gnn_fraud_detection_pipeline` which in turn requires the user to have pipe installed in the Docker container.

docker/Dockerfile
pdmack: @mdemoret-nv conda silently ignores the removal. Probably internally "protected".

pdmack: @mdemoret-nv you're right. Moved down to runtime stage. PTAL

pdmack: @dagardner-nv done. PTAL

mdemoret-nv (approved):

PR #1062, opened 2023-07-18
Fix phishing `config.pbtxt` which was causing performance degradation

efajardo-nv:
## Description
- Phishing E2E benchmarks revealed 4x throughput degradation from previous benchmarks caused by accidental change to `config.pbtxt`.  `precision_mode` was changed from `FP16` to `FP32`.
- Reverted the above change and now showing throughput improvement from previous release benchmark.

## Checklist
[x] I am familiar with the [Contributing Guidelines](https://github.com/nv-morpheus/Morpheus/blob/main/docs/source/developer_guide/contributing.md).
[x] New or existing tests cover these changes.
[x] The documentation is up to date with these changes.


mdemoret-nv (approved):
Nice catch

PR #1115, opened 2023-08-02
Break circular reference issue causing a memory leak

### LINKED ISSUE #1114: [BUG]: Stage destructors not called
### Version

23.11

Source

### Describe the bug.

I noticed that the destructor for a C++ stage isn't called, even when the pipeline has gone out of scope, and `gc.collect()` has been called.

```shell
import gc
import logging
import os
import shutil
import tempfile

from morpheus.utils.logger import configure_logging
from morpheus.config import Config
from morpheus.config import CppConfig
from morpheus.pipeline import LinearPipeline
from morpheus.stages.input.file_source_stage import FileSourceStage
from morpheus.stages.output.write_to_file_stage import WriteToFileStage

MORPHEUS_ROOT_DIR = os.environ['MORPHEUS_ROOT']
TEST_DF = os.path.join(MORPHEUS_ROOT_DIR, "tests", "tests_data", "filter_probs.csv")


def run_pipe(out_file: str):
    config = Config()
    pipe = LinearPipeline(config)
    pipe.set_source(FileSourceStage(config, filename=TEST_DF))
    pipe.add_stage(WriteToFileStage(config, filename=out_file, overwrite=False))
    pipe.run()


def wrapper(logger: logging.Logger):
    tmp_dir = tempfile.mkdtemp()
    out_file = os.path.join(tmp_dir, "test.csv")

    logger.debug("Running pipe")
    try:
        run_pipe(out_file)
    finally:
        logger.debug("Pipe finished, cleaning up")
        shutil.rmtree(tmp_dir)
        logger.debug("Calling gc.collect()")
        gc.collect()
        logger.debug("gc.collect() - done")


if __name__ == "__main__":
    configure_logging(log_level=logging.DEBUG)
    logger = logging.getLogger("morpheus.tests.test_destructors")
    #CppConfig.set_should_use_cpp(False)
    # gc.set_debug(gc.DEBUG_LEAK)
    wrapper(logger)
```

### Relevant log output

_No response_

### Full env printout

_No response_

### Other/Misc.

_No response_

### Code of Conduct

dagardner-nv:
## Description
* Fixes a memory leak #1114 by releasing reference to MRC Pipeline & Executor on stop.
* Numerous pylint fixes for `morpheus/pipeline/pipeline.py`
* Skip `tests/test_multi_segment.py::test_multi_segment_bad_data_type` due to nv-morpheus/MRC#360

fixes #1114

## Checklist


morpheus/pipeline/pipeline.py
```diff
@@ -232,12 +228,16 @@ def build(self):
...
+        exec_options.engine_factories.default_engine_type = mrc.core.options.EngineType.Thread
+
+        self._mrc_executor = mrc.Executor(exec_options)
 
-        self._mrc_pipeline = mrc.Pipeline()
+        mrc_pipeline = mrc.Pipeline()
 
         def inner_build(builder: mrc.Builder, segment_id: str):
-            logger.info(f"====Building Segment: {segment_id}====")
+            logger.info("====Building Segment: %s ====", segment_id)
```
cwharris, line 240: How come you are switching to this format when in most places you are switching to the other?

tests/test_multi_segment.py
```diff
@@ -36,6 +36,7 @@ def test_linear_boundary_stages(config, filter_probs_df):
     assert_results(comp_stage.get_results())
 
 
+@pytest.mark.skip(reason="Skipping due to MRC issue #360")
```
cwharris, line 39: Is this something we can `xfail` instead? That way we can notified when it starts passing, which it would do if we fix the bug in MRC, but otherwise not notice it was fixed in Morpheus.

cwharris (approved):
LGTM with comments

morpheus/pipeline/pipeline.py
dagardner-nv: This is a pylint recommendation. For logger calls the string interpolation isn't performed when the configured log-level would cause the call to be suppressed. 

For other strings, pylint recommends the new-ish f-style strings.

tests/test_multi_segment.py
dagardner-nv: In this case we cannot as marking a test `xfail` still executes the test, and this test would abort the current process.

morpheus/io/data_manager.py
```diff
@@ -49,6 +49,11 @@ def __init__(self, storage_type: str = 'in_memory', file_format: str = 'parquet'
         :param file_format: Specifies the file format to be used. Can be either 'parquet' or 'csv'.
         """
 
+        # Define these early so that they are defined even if we raise an exception, this ensures that we don't get an
+        # attribute error in the __del__ method.
+        self._storage_dir = None
+        self._storage_type = None
+
```
dagardner-nv, line 56: Unrelated fix, just noticed it.

tests/conftest.py
```diff
@@ -561,6 +562,15 @@ def reset_plugins(reset_plugin_manger, reset_global_stage_registry):
     yield
 
 
+@pytest.fixture(scope="function")
+def disable_gc():
```
mdemoret-nv, line 566: Add a comment describing this fixture and when to use it

tests/test_gc_tls.py
```diff
@@ -0,0 +1,98 @@
...
+    pipeline has finished running. Then run a second pipeline in either C++ or Python mode, where once of the stages
+    has a finalizer on the thread which calls `gc.collect` specifically `cupy.fft` does this and I was unable to repro
+    the issue myself, possibly since that code was written in cython.
+
+    This only works if the first pipeline and thus the `PyDataTable` are not gargabe collected before the second
+    pipeline is collected which makes this difficult to reproduce, but it is possible to force the issue by disabling
+    automatic garbage collection.
+    """
+    run_pipe1(filter_probs_df)
+    run_pipe2(filter_probs_df)
```
mdemoret-nv, line 98: Is this needed anymore now that we better understand the issue on the MRC side?

tests/test_multi_segment.py
mdemoret-nv: @dagardner-nv Can we remove this once we are using the fix in MRC?

tests/test_pipe.py
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python
```
mdemoret-nv, line 1: Rename this to something like "test_pipeline.py"

tests/test_pipe.py
```diff
@@ -0,0 +1,83 @@
...
+        self._state_dict = state_dict
+        self._state_key = state_key
+
+    @property
+    def name(self) -> str:
+        return "test-sink"
+
+    def __del__(self):
+        self._state_dict[self._state_key] = True
+        self._state_dict = None
```
mdemoret-nv, line 63: The way these classes are organized could easily break if the `state_dict` is accidentally copied instead of referenced. I think a better organization would be to take a lambda function which gets executed in `__del__`. This way, the lambda function can use `nonlocal` to ensure we dont copy the `state_dict`. Something like:

```python
state_dict = {"source": False, "sink": False}

def update_state_dict(key:str):
   nonlocal state_dict
   state_dict[key] = True

pipe.set_source(SourceTestStage(config, [filter_probs_df], lambda: update_state_dict("source"))
pipe.set_source(SinkTestStage(config, lambda: update_state_dict("sink"))
```

mdemoret-nv (approved):
Approving. Merge once comments are resolved. I think we can remove some of the GC tests since they can be executed better in MRC

PR #1118, opened 2023-08-03
Fix Tests broken by Test Reorganization

cwharris:
The test reorganization resulted in a couple of problems. First, not all tests were being run due to a mistake I made in handling cmake function arguments. Second, the `test_cuda` test was producing a segfault inside of triton client while initializing curl, but only in debug builds. The fix was to the segfault was not obvious, but apparently only occured when the test utilities were compiled as a separate library. Interestingly, the other tests were not affected. Perhaps there is some global state that needs to be managed that is not managed in the simple `test_cuda` test, which only calls a Matx function (also important, the Matx function result is not inspected, so the test itself is not comprehensive, as Matx does not produce an error if the kernel is not available for the given architecture).

Depends on https://github.com/nv-morpheus/Morpheus/pull/1110


mdemoret-nv:
Quick update: Were having issues with PR #1110 which is a blocker for this PR. Is there any way to fix this without that PR to allow dev testing locally again?

mdemoret-nv (approved):

mdemoret-nv:
@cwharris Due to some unrelated issues, we likely wont merge #1110 as it is. Can you remove those changes from this PR so it can get merged?

cwharris:
@mdemoret-nv done.

PR #1032, opened 2023-07-11
Update GNN  stellargraph with DGL

tzemicheal:
## Description

This PR replaces stellargraph reference in the training & validation scripts. This will address the issue #945  & #925

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/nv-morpheus/Morpheus/blob/main/docs/source/developer_guide/contributing.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #1135, opened 2023-08-18
Prefer conda package over pip dependencies

### LINKED ISSUE #1129: [FEA]: Prefer conda packages and remove as many pip packages as possible from conda environment

Change

Medium

We have several pip packages being installed in our conda environment which have matching conda packages. Or they have dependencies which could be installed by conda. In almost every situation, we should prefer conda over pip in our conda environment config files.

The main driver for this is if a package is installed by pip and the same package but a different version is installed by conda, then the two packages collide and neither works correctly. Since pip does not do any dependency resolution, we should prefer conda (even with all of its problems).

This issue should be resolved by performing the following:

- Search all packages being installed by pip and check if there is a matching conda package. If so, move it to the conda environment. 
- Any packages which cannot be installed by conda, check if there are any major dependencies which could be moved to the conda environment
   - This can be verified by performing a dry run of a pip install and see what packages it would have installed. Check to see if there are conda packages of those.

In the end, the number of packages that are installed by pip should be **_very low_**.

_No response_

### Additional context

_No response_

### Code of Conduct

cwharris:
Prefer conda package over pip dependencies

Todo:
- [x] build and test inside devcontainer
- [x] build and test inside development container
- [ ] build and test inside release container

Closes #1129


ci/runner/Dockerfile
```diff
@@ -67,6 +67,7 @@ COPY ./docker/conda/environments/cuda${CUDA_SHORT_VER}_examples.yml /tmp/conda/c
 COPY ./ci/scripts/download_kafka.py /tmp/scripts/download_kafka.py
 
 # Install extra deps needed for gnn_fraud_detection_pipeline & ransomware_detection examples
+# can stellargraph be obtained through conda?
```
mdemoret-nv, line 70: Is this comment intentional?

docker/conda/environments/cuda11.8_dev.yml
```diff
@@ -82,24 +86,18 @@ dependencies:
...
+    - tritonclient=2.26 # Required by NvTabular, force the version, so we get protobufs compatible with 4.21
     - typing_utils=0.1
     - watchdog=2.1
+    - websockets
     - yapf=0.40.1
-    ####### Morpheus Pip Dependencies (keep sorted!) #######
-    - pip:
-        # Add additional dev dependencies here
-        - docutils
-        - pytest-kafka==0.6.0
```
mdemoret-nv, line 103: What happened to `pytest-kafka`?

mdemoret-nv (commented):
Couple of questions.

ci/runner/Dockerfile
cwharris: I saw we were using pip to install stellargraph here and left the comment so I wouldn't forget to check on it later. Guess it worked, in a round-about way. Should we try to install stellargraph via conda and get rid of this pip install call?

docker/conda/environments/cuda11.8_dev.yml
cwharris: Good question. CI passes. Does it need to be installed in release? There's a script to install it in the devcontainer.

ci/runner/Dockerfile
mdemoret-nv: Lets leave it alone for now. There is a PR: #1032 which will remove stellargraph entirely. No need to worry about it for now.

docker/conda/environments/cuda11.8_dev.yml
mdemoret-nv: `pytest-kafka` is needed for running the tests. We install it manually in the Dockerfile to ensure its setup for CI but this doesnt cover normal developer use. I think it should be left in the dev dependencies.

cwharris:
@mdemoret-nv ready for re-review

mdemoret-nv (approved):

PR #1155, opened 2023-08-24
Update dfp_training stage to support ControlMessages or MultiDFPMessages

### LINKED ISSUE #1148: [FEA]: Update `DFPTraining` stage to accept both `ControlMessage` and `MultiDFPMessage`

New Feature

High

Currently, the DFP pipeline is implemented using both Modules and Stages. The modules exclusively work with `ControlMessage` while the stages work exclusively with `MultiDFPMessage`. This makes it difficult for users customizing the DFP pipeline to use some of the advantages of `ControlMessage` if they are using stages. One particular area where `ControlMessage` would be more beneficial is the training stage to allow using data loaders. 

As a first step, the `DFPTraining` stage should be updated to accept both `ControlMessage` and `MultiDFPMessage` and work with either. If it accepts a `ControlMessage` it should return a `ControlMessage` and the same for `MultiDFPMessage`.

Completion Criteria:

- [x] The `DFPTraining` accepts both `ControlMessage` and `MultiDFPMessage`
   - [x] The type can be determined in the build phase and remain static
- [x] The `DFPTraining` returns a ControlMessage if one is passed in or a MultiAEMessage if a DFPMessageMeta is passed in
- [ ] The `DFPTraining` stage should always pass the data to `AutoEncoder` as a data loader (even if its a `MultiDFPMessage`)
- [x] The training results should be unchanged

_No response_

### Additional context

_No response_

### Code of Conduct

drobison00:
## Description
Closes #1148

This addresses relevant issues with 1148, additionally, in order to make the .fit interface work in a more unified way, we've merged and simplified the _fit_centralized and _fit_distributed code paths, and eliminated large portions of dead code.

## Checklist
- [x ] I am familiar with the [Contributing Guidelines](https://github.com/nv-morpheus/Morpheus/blob/main/docs/source/developer_guide/contributing.md).
- [x ] New or existing tests cover these changes.
- [x ] The documentation is up to date with these changes.


morpheus/models/dfencoder/autoencoder.py
```diff
@@ -832,97 +835,79 @@ def _fit_centralized(self, df, epochs=1, val=None, run_validation=False, use_val
             raise ValueError("Validation set is required if either run_validation or \
                 use_val_for_loss_stats is set to True.")
 
-        if use_val_for_loss_stats:
-            df_for_loss_stats = val.copy()
-        else:
-            # use train loss
-            df_for_loss_stats = df.copy()
+        val_dset = None
+        if run_validation:
```
drobison00, line 839: May need to copy DF

morpheus/models/dfencoder/autoencoder.py
```diff
@@ -1010,112 +868,94 @@ def _fit_distributed(
...
 
-                train_loss_count += 1
-                train_loss_sum += loss
+        if (is_main_process):
+            loss_dset.convert_to_validation(self)
+            self._populate_loss_stats_from_dataset(loss_dset)
 
-            if self.lr_decay is not None:
-                self.lr_decay.step()
+    def _validate_dataframe(self, orig_df, swapped_df):
```
hsin-c, line 913: We don't seem to be using this function anymore, should we remove it altogether?

examples/digital_fingerprinting/production/morpheus/dfp/stages/dfp_training.py
```diff
@@ -114,15 +148,23 @@ def on_data(self, message: MultiDFPMessage) -> MultiAEMessage:
         model.fit(train_df, epochs=self._epochs, val_data=validation_df, run_validation=run_validation)
         logger.debug("Training AE model for user: '%s'... Complete.", user_id)
 
-        output_message = MultiAEMessage(meta=message.meta,
-                                        mess_offset=message.mess_offset,
-                                        mess_count=message.mess_count,
-                                        model=model)
+        if (to_cm):
+            output_message = ControlMessage(message.meta)
+            output_message.set_metadata("user_id", user_id)
```
mdemoret-nv, line 153: What happens to the model here? Is it just lost?

morpheus/models/dfencoder/dataloader.py
```diff
@@ -352,89 +349,46 @@ def get_preloaded_data(self):
...
+    @property
+    def batch_size(self):
+        return self._batch_size
 
-    @staticmethod
-    def get_validation_dataset(model, data_folder, load_data_fn=pd.read_csv, preload_data_into_memory=True):
-        """A helper function to get a validation dataset with the provided parameters.
+    @batch_size.setter
+    def batch_size(self, value):
+        self._batch_size = value
```
mdemoret-nv, line 366: I feel like you shouldnt be able to change the batch size after the dataset has been created? If its in the middle of creating batches, this could cause a problem, right?

morpheus/models/dfencoder/dataloader.py
```diff
@@ -352,89 +349,46 @@ def get_preloaded_data(self):
...
-            model.eval_batch_size,
-            model.preprocess_validation_data,
-            load_data_fn=load_data_fn,
-            shuffle_rows_in_batch=False,
-            preload_data_into_memory=preload_data_into_memory,
-        )
-        return dataset
+    @shuffle_rows_in_batch.setter
+    def shuffle_rows_in_batch(self, value):
+        self._shuffle_rows_in_batch = value
```
mdemoret-nv, line 374: Again, do we need the setter for this property?

morpheus/models/dfencoder/dataloader.py
```diff
@@ -545,65 +499,34 @@ def _preprocess(self, df, batch_index):
...
-         model : AutoEncoder
-            The autoencoder model used to get relevant params and the preprocessing func.
-        """
-        self._preprocess_fn = model.preprocess_validation_data
-        self._batch_size = model.eval_batch_size
-        self._shuffle_rows_in_batch = False
-        self._shuffle_batch_indices = False
+    @shuffle_batch_indices.setter
+    def shuffle_batch_indices(self, value):
+        self._shuffle_batch_indices = value
```
mdemoret-nv, line 532: These are the same between the two dataset classes in this file. Should they be abstracted to a common base class?

morpheus/models/dfencoder/autoencoder.py
```diff
@@ -737,251 +741,132 @@ def _create_stat_dict(self, a):
...
+
+        return dataset
+
+    def _transform_dataset_for_validation(self, dataset):
+        dataset.batch_size = self.eval_batch_size
+        dataset.preprocess_fn = self.preprocess_validation_data
+        dataset.shuffle_batch_indices = False
+        dataset.shuffle_rows_in_batch = False
+
+        return dataset
```
mdemoret-nv, line 758: Ok I see why you need these setters on the dataset classes. I still feel that it would be better to set these in the constructor since they are used in `__iter__`. If you changed these properties between iterations, you could introduce bugs

mdemoret-nv (changes requested):
- [x] Audit the property setters on `FileSystemDataset`. Seems like changing some of the properties while the dataset is iterating might be an issue
- [ ] Add tests to execute some of the new checks added to `AutoEncoder` (such as the validation on input types, and any calls that can raise an exception)

morpheus/models/dfencoder/dataloader.py
drobison00: Yes, although we don't make any kind of contract that this is thread safe, so if someone is changing batch size while iterating that seems like an algorithmic problem; we also do change the batch size and various other config params when we transition from doing training to the validation step.

morpheus/models/dfencoder/dataloader.py
drobison00: We don't need a 'setter' necessarily, but we do need to be able to set it; the use case is when transitioning from training to validation.

morpheus/models/dfencoder/dataloader.py
drobison00: Hsin and I talked about it; since they're specific to DFP, and the classes themselves are fairly distinct in terms of function, I'm kind of indifferent. I also expect this will change or go away entirely when we transition to ControlMessages with managed payloads.

morpheus/models/dfencoder/autoencoder.py
drobison00: I agree, but anyone changing things between iterations is making a conscious decision to do so, I don't necessarily think we should avoid these transforms as they save us copying the dataset otherwise.

examples/digital_fingerprinting/production/morpheus/dfp/stages/dfp_training.py
drobison00: In this case, yes; we were not capturing the model in the ControlMessage paths. Will take a look.

drobison00: Serializing the model and storing it in the payload metadata in the next update.

morpheus/models/dfencoder/dataloader.py
mdemoret-nv: Fair enough.

morpheus/models/dfencoder/autoencoder.py
mdemoret-nv: Fair enough

cwharris (changes requested):
Kind of amazed at the amount of code reduction in this PR.

cwharris:
Do we have tests that are going to cover these changes? I don't see any tests explicitly for AutoEncoder.

examples/digital_fingerprinting/production/morpheus/dfp/stages/dfp_training.py
```diff
@@ -87,10 +89,42 @@ def supports_cpp_node(self):
...
-        return (MultiDFPMessage, )
+        return (
+            ControlMessage,
+            MultiDFPMessage,
+        )
 
+    def _dfp_mm_from_cm(self, control_message: ControlMessage) -> MultiDFPMessage:
+        """Create a MultiDFPMessage from a ControlMessage."""
+        user_id = control_message.get_metadata("user_id")
+        message_meta = control_message.payload()
```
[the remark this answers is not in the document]
drobison00: The general idea is metadata is a place to store arbitrary configuration items, while the payload is something common to every ControlMessage.

Another change coming with 23.11 is that the payload is going to be a more intelligent data storage object, one that's capable of dynamically caching its contents to make its use more or less independent of the size of its data. To start this will just be storage in system memory, or caching to disk, but also supporting iterating using torch datasets.

morpheus/models/dfencoder/dataloader.py
```diff
@@ -41,10 +41,7 @@ def __iter__(self):
             if self.batch_size == 1:
                 # unbatch to get rid of the first dimention of 1 intorduced by DataLoaders batching
                 # (if batch size is set to 1)
-                data_d["data"] = {
-                    k: v[0] if type(v) != list else [_v[0] for _v in v]
-                    for k, v in data_d["data"].items()
-                }
+                data_d["data"] = {k: v[0] if type(v) != list else [_v[0] for _v in v] for k, v in data_d["data"].items()}
```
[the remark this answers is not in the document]
drobison00: Yeah, I generally agree. Because I'm expecting a lot of changes or even having some of these objects go away in the future for a fully general approach, I'm inclined to leave this as is.

morpheus/models/dfencoder/autoencoder.py
```diff
@@ -737,251 +741,132 @@ def _create_stat_dict(self, a):
...
-            Data used for training.
-        epochs : int, optional
-            Number of epochs to run training, by default 1.
-        val : pandas.DataFrame, optional
-            Optional pandas dataframe for validation or loss stats, by default None.
-        run_validation : bool, optional
-            Whether to collect validation loss for each epoch during training, by default False.
-        use_val_for_loss_stats : bool, optional
-            Whether to use the validation set for loss statistics collection (for z score calculation), by default False.
+        return dset if not is_loader else data
```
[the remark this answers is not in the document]
drobison00: We need to transform the dataset whether its created by us, passed by a user, or passed inside of a DataLoader, because we modify it for training vs validation and want to be sure everything is in a known consistent state prior to training.

I'll add a comment.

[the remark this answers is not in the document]
cwharris: So true transformations are mutative. That makes sense. It’s just not immediately obvious since we’re also sometimes returning a value.

morpheus/models/dfencoder/autoencoder.py
```diff
@@ -737,251 +741,132 @@ def _create_stat_dict(self, a):
...
-            if self.n_megabatches > 1:
-                self.train_megabatch_epoch(n_updates, df)
-            else:
-                input_df = df.swap(likelihood=self.swap_p)
-                self.train_epoch(n_updates, input_df, df)
+                train_loss_count += 1
+                train_loss_sum += loss
 
-            if self.lr_decay is not None:
+            if (self.lr_decay is not None):
```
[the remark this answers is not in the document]
drobison00: It stands for 'learning rate' decay. I don't think it would be confusing from a DS perspective, but I also don't see any reason not to make it's function more clear, will update.

mdemoret-nv (approved):

cwharris (approved):

PR #1160, opened 2023-08-29
Scope Zookeeper & Kafka fixtures to session

dagardner-nv:
## Description
* `pytest --run_slow --fail_missing --run_kafka` 2 minutes faster
* `kafka_consumer` & `kafka_topics` fixtures are still function scoped, ensuring each test will receive unique topics & client
* Move Kafka related testing code to `tests/_utils/kafka.py`

## Checklist


tests/_utils/kafka.py
```diff
@@ -0,0 +1,152 @@
...
+PARTITION_ASSIGNMENT_TIMEOUT = 15
+
+# actual topic names not important, but we will need two of them.
+KafkaTopics = namedtuple('KafkaTopics', ['input_topic', 'output_topic'])
+
+# pylint: disable=invalid-name
+_kafka_consumer = None
+kafka_server = None
+zookeeper_proc = None
+# pylint: enable=invalid-name
```
cwharris, line 39: what's with the pylint directives? aside from the `_kafka_consumer`, these look pretty normal?

tests/_utils/kafka.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+@pytest.fixture(name='kafka_topics', scope='function')
+def kafka_topics_fixture():
+    yield KafkaTopics(f'morpheus_unittest_input_{time.time()}', f'morpheus_unittest_output_{time.time()}')
```
cwharris, line 44: Is `time` the best discriminator we can use? This has the potential to create a lot of topics if we run tests repeatedly without restarting kafka. What are your thoughts on using something like this?
```python
import os
import re
def get_test_name_for_kafka_topic():
    test_name = os.environ.get("PYTEST_CURRENT_TEST")
    test_name = re.sub("[:/\[\]\(\)]", "-", test_name)
    test_name = re.sub(",", "_", test_name)
    return test_name
```
which produces something along these lines:
```python
"tests-test_kafka_oauth.py--test_kafka_oauth-use_cpp- -call-"
"tests-test_kafka_oauth.py--test_kafka_oauth-use_python- -call-"
```

tests/_utils/kafka.py
```diff
@@ -0,0 +1,152 @@
...
+    producer = KafkaProducer(bootstrap_servers=bootstrap_servers, client_id=client_id)
+    for row in data:
+        if isinstance(row, dict):
+            row = json.dumps(row)
+        producer.send(kafka_topic, row.encode('utf-8'))
+        num_records += 1
+
+    producer.flush()
+
+    assert num_records > 0
```
cwharris, line 133: Why are we summing `num_records`? Wouldn't that be the same as `len(data)`? Unless `producer.send` or `json.dumps` raises an error... in which case we wouldn't use the value anyway?

tests/_utils/kafka.py
```diff
@@ -0,0 +1,152 @@
...
+    return num_records
+
+
+def write_file_to_kafka(bootstrap_servers: str,
+                        kafka_topic: str,
+                        input_file: str,
+                        client_id: str = 'morpheus_unittest_writer') -> int:
+    """
+    Writes data from `inpute_file` into a given Kafka topic, emitting one message for each line int he file.
+    Returning the number of messages written
```
cwharris, line 144:
```suggestion
    Writes data from `input_file` into a given Kafka topic, emitting one message for each line in the file, 
    and returns the number of messages written.
```

cwharris (approved):
Thanks for consolidating all of the kafka utils to the appropriate file. :pray:

tests/_utils/kafka.py
dagardner-nv: Pylint believes these to be constants and wants an upper-case name.

tests/_utils/kafka.py
dagardner-nv: I don't recall how we ended up with this code. I didn't modify this method, simply moved it from another module, however I believe you are correct num_records should be the same as `len(data)`

tests/_utils/kafka.py
cwharris: weird!

mdemoret-nv (approved):

PR #1170, opened 2023-09-01
Multiprocess Stage POC

cwharris:
Example demonstrates the use of Python Multiprocessing with a custom Stage, which enables offloading of MultiMessage processing to a separate process while continuing the workflow within the same process.

The example results in a segfault originating from `Watchable::watcher_epilogue` when calling `sub.on_next` from python in the parent process. It appears that one of `m_watchers` has been released and we're still trying to access it. Unfortunately I don't know enough about MRC to say what the cause is.

```c++
inline void Watchable::watcher_epilogue(WatchableEvent op, bool rc, const void* addr)
{
    for (const auto& obs : m_watchers)
    {
        obs->on_exit(op, rc, addr); // <- segfault occurs here, likely when de-referencing `obs`
    }
}
```
I was able to work around the issue by simply commenting out the `obs->on_exit(...)` call and rebuilding MRC and Morpheus together. Everything seemed to work fine, so maybe this is a bug in MRC?

## Checklist


jarmak-nv:
Hey @cwharris should we promote this out of draft since you're requested review?

Salonijain27:
Closing this PR as it has been open for over a year. If required please reopen it.

PR #1284, opened 2023-10-17
Skip milvus tests by default

dagardner-nv:
## Description
* Builds on PR #1273
* Skip milvus tests by default requiring a `--run_milvus` flag, similar to the kafka tests which require the `--run_kafka` flag.
* Milvus takes over a minute just to start up, and the tests require an additional three minutes. Requiring four minutes just to run the milvus tests alone, while the rest of the slow tests take only a minute and a half.
* Consolidate milvus related fixtures.
* Specify an explicit temp dir as the storage location for milvus, ensuring we always start with an empty database.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -64,6 +64,13 @@ def pytest_addoption(parser: pytest.Parser):
     )
 
     parser.addoption(
+        "--run_milvus",
+        action="store_true",
+        dest="run_milvus",
+        help="Run milvus tests that would otherwise be skipped",
+    )
+
+    parser.addoption(
         "--run_benchmark",
         action="store_true",
         dest="run_benchmark",
@@ -145,6 +152,10 @@ def pytest_runtest_setup(item):
     if (not item.config.getoption("--run_kafka")):
         if (item.get_closest_marker("kafka") is not None):
             pytest.skip("Skipping Kafka tests by default. Use --run_kafka to enable")
+
+    if (not item.config.getoption("--run_milvus")):
+        if (item.get_closest_marker("milvus") is not None):
+            pytest.skip("Skipping milvus tests by default. Use --run_milvus to enable")
 
     if (not item.config.getoption("--run_benchmark")):
         if (item.get_closest_marker("benchmark") is not None):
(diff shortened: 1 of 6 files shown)


tests/conftest.py
```diff
@@ -972,11 +982,30 @@ def milvus_server_uri():
...
-            logger.error("Error in stopping Milvus server: %s", exec_inf)
+
+
+@pytest.fixture(scope="function", name="milvus_service")
+def milvus_service_fixture(milvus_server_uri: str):
+    # This fixture is scoped to the function level since the WriteToVectorDBStage will close the connection on'
+    # pipeline completion
+    from morpheus.service.milvus_vector_db_service import MilvusVectorDBService
+    service = MilvusVectorDBService(uri=milvus_server_uri)
+    yield service
```
mdemoret-nv, line 993: This seems like it should be something scoped to the milvus tests. The conf tests should be free of objects we are testing

tests/conftest.py
```diff
@@ -958,12 +969,11 @@ def milvus_server_uri():
     from milvus import default_server
 
     logger = logging.getLogger(f"morpheus.{__name__}")
-    try:
-        # Milvus checks for already bound ports but it doesnt seem to work for webservice_port. Use a random one
-        default_server.webservice_port = _get_random_port()
 
-        # Now start the server
-        default_server.start()
```
mdemoret-nv, line 966: What happened to starting the server?

mdemoret-nv (approved):
Good changes, but it seems like we should copy the camoflague stuff even more. i.e. use an environment variable to prevent starting a milvus server (so you can run it yourself). And only launch the standalone service when needed

tests/conftest.py
dagardner-nv: Still happens but we are using `default_server` as a context manager.

tests/conftest.py
dagardner-nv: This fixture is used by both `tests/test_milvus_vector_db_service.py` and by `tests/test_milvus_write_to_vector_db_stage_pipe.py`. 

As far as I'm aware the only way to allow them to share the fixture is to either place them here, or move the two milvus test to their own sub-directory.

PR #1294, opened 2023-10-20
Improvements to MilvusVectorDB Service

### LINKED ISSUE #1272: [FEA]: Improve the `VectorDBService` and `WriteToVectorDBStage` to handle similarity search

Improvement

High

When working with the `VectorDBService` and `WriteToVectorDBStage` for LLM pipelines, there are a few deficiencies which have come up:

- There is no API for performing a similarity search from a `FLOAT_VECTOR` index type
- The resource name needs to be specified for every call to the service API.
- No information can be saved in the service for a particular resource (i.e. the schema or configuration options)
- The `WriteToVectorDBStage` does not have any capability to upload embedding vectors

The following improvements should be added to the `VectorDBService` class:

1. A new sub object (`VectorDBServiceResource`) can be created which limits the API to a single resource.
   1. All of the APIs which take a `name` parameter should be available on the `VectorDBServiceResource` class.
   1. i.e. `VectorDBService.insert(name: str, data: list, **kwargs)` -> `VectorDBServiceResource.insert(data: list, **kwargs)`
1. A new method `get_resource_client(self, name: str, **kwargs: dict[str, typing.Any]) -> VectorDBServiceResource` will provide the mechanism for getting the resource client object
1. Reorganizing the files:
   1. Make a subdirectory in `morpheus/service` called `vdb`
   1. Move `milvus_vector_db_service.py` to `morpheus/service/vdb`
   1. Move `vector_db_service.py` to `morpheus/service/vdb`
   1. Move `vector_db_service_utils.py` to `morpheus/service/vdb/utils.py`

These will then need to be implemented for all VDB implementations.

The following improvements should be added to the `WriteToVectorDBStage`:

1. Ability to specify which column should be treated as an embedding vector column for the purpose of uploading an embedding with each document
1. The ability to recreate the VDB resource on startup
1. Separation of the `service_kwargs` (the arguments needed for connecting to the service) from the `resource_kwargs` (the args required to build and setup the resource)
1. The `service_kwargs` will always be required since they are necessary to connect to the service
1. The `resource_kwargs` are only necessary when creating the resource (or recreating). If we are just connecting to an existing resource, then the parameters can be determined from the service.
1. Ability to use both `ControlMessage` or `MultiMessage` or `MultiMessageResponse` (if an embedding column is specified) types

### Additional context

See the following files for partial implementations of the above:

- https://github.com/nv-morpheus/Morpheus/blob/de35ac300e7061bcfd3599c2c7f9b3373e71d703/morpheus/stages/output/write_to_vector_db.py
- https://github.com/nv-morpheus/Morpheus/blob/de35ac300e7061bcfd3599c2c7f9b3373e71d703/morpheus/service/milvus_vector_db_service.py
- https://github.com/nv-morpheus/Morpheus/blob/b324ed725c51d38f8b8437dbfbd3d55788b5d2a3/morpheus/service/vector_db_service.py
- https://github.com/nv-morpheus/Morpheus/blob/b324ed725c51d38f8b8437dbfbd3d55788b5d2a3/morpheus/utils/vector_db_service_utils.py

### Code of Conduct

bsuryadevara:
## Description
- Updated MilvusVectorDBService Implementation to handle similarity search. 
Closes #1272 


morpheus/service/vdb/milvus_vector_db_service.py
```diff
@@ -217,6 +384,158 @@ async def similarity_search(self,
...
+        **kwargs :  dict[str, typing.Any]
+            Extra keyword arguments specific to the vector database implementation.
+
+        Returns
+        -------
+        typing.Any
+            Returns result of the given keys that are deleted from the collection.
+        """
+
+        result = self._client.delete(collection_name=self._name, pks=keys, **kwargs)
```
drobison00, line 430: Just want to be sure this is correct? Other patterns appear to do something like:

```python
self._command
self._flush
```

morpheus/service/vdb/milvus_vector_db_service.py
```diff
@@ -217,6 +384,158 @@ async def similarity_search(self,
...
+
+        return result
+
+    def delete(self, expr: str, **kwargs: dict[str, typing.Any]) -> dict[str, typing.Any]:
+        """
+        Delete vectors from the collection using expressions.
+
+        Parameters
+        ----------
+        expr : str
```
drobison00, line 442: Is this just like a regex? if its something custom, can you put a link in the doc string.

morpheus/service/vdb/milvus_vector_db_service.py
```diff
@@ -217,6 +384,158 @@ async def similarity_search(self,
...
+
+        This function allows you to drop a collection.
+
+        Parameters
+        ----------
+        **kwargs : dict
+            Additional keyword arguments for specifying the type and partition name (if applicable).
+        """
+
+        self._collection.drop(**kwargs)
```
drobison00, line 512: Does this require an associated flush?

drobison00 (approved):
Approved; one or two small comments.

morpheus/controllers/rss_controller.py
```diff
@@ -153,20 +180,17 @@ def _try_parse_feed(self, url: str) -> feedparser.FeedParserDict:
...
-                    logger.warning("Failed to parse feed: %s, %s. Trying with other source",
-                                   url,
-                                   feed['bozo_exception'])
+                    logger.info("Failed to parse feed: %s, %s. Try parsing the feed manually",
+                                url,
+                                feed['bozo_exception'])
                     feed = self._try_parse_feed_with_beautiful_soup(url, is_url)
-                except Exception as exec_info:
-                    raise RuntimeError(f"Invalid feed input: {url}. Error: {exec_info}") from exec_info
+                except Exception:
```
drobison00, line 192: Does it make sense to add a logging error here?

morpheus/controllers/rss_controller.py
```diff
@@ -222,45 +254,22 @@ def fetch_dataframes(self):
...
-                            yield self.create_dataframe(entry_accumulator)
+                            yield cudf.DataFrame(entry_accumulator)
                             entry_accumulator.clear()
 
             self._previous_entries = current_entries
 
             # Yield any remaining entries.
             if entry_accumulator:
-                df = self.create_dataframe(entry_accumulator)
+                # TODO (Bhargav): Debug : cannot mix list and non-list, non-null values error
```
drobison00, line 264: TODO

tests/controllers/test_rss_controller.py
```diff
@@ -142,9 +136,52 @@ def test_try_parse_feed_with_beautiful_soup(feed_input: str, is_url: bool, enabl
...
     if enable_cache:
-        assert rss_controller._session
+        assert controller._session
     else:
-        assert not rss_controller._session
+        assert not controller._session
+
+
+def test_parse_feeds():
+    feed_input = "https://rss.nytimes.com/services/xml/rss/nyt/HomePage.xml"
```
drobison00, line 148: I'm not sure I like the idea of hitting an external site in our unit tests, can we just capture one of these calls and store the data for use in the parsing test?

drobison00 (changes requested):

morpheus/service/vdb/milvus_vector_db_service.py
bsuryadevara: It's not exactly a regex, but it is similar as shown below.
`expr = "book_id in [0,1]"`

here is the doc: https://milvus.io/docs/delete_data.md

morpheus/service/vdb/milvus_vector_db_service.py
bsuryadevara: Drop doesn't need flush, as it is deleting the collection.

morpheus/service/vdb/milvus_vector_db_service.py
bsuryadevara: Delete doesn't need flush command. They are automatically getting updated in the collection. When I do search operation on the collection after delete. It is retrieving only keys that are not deleted.

We have a tests covering this scenario:
https://github.com/nv-morpheus/Morpheus/blob/b67882a6f62c304a6eb6260fa1d228e7e5ada2dc/tests/test_milvus_vector_db_service.py#L309
https://github.com/nv-morpheus/Morpheus/blob/b67882a6f62c304a6eb6260fa1d228e7e5ada2dc/tests/test_milvus_vector_db_service.py#L333

tests/controllers/test_rss_controller.py
bsuryadevara: Removed external sites.

morpheus/controllers/rss_controller.py
```diff
@@ -182,17 +206,25 @@ def parse_feeds(self):
...
+                    feed_stats.success_count += 1
+                    feed_stats.last_try_result = "Success"
 
-                yield feed
+                    yield feed
 
             except Exception as ex:
-                logger.warning("Failed to parse feed: %s: %s. The feed will be not be retried.", url, ex)
-                self._errored_feeds.append(url)
+                logger.info("Failed to parse feed: %s: %s.", url, ex)
```
drobison00, line 222: I think leaving this as a warning is probably the right answer vs info, anyone monitoring the pipeline's output will probably want to see this.

Could you also update the string to include the number of failures and total attempts.

tests/controllers/test_rss_controller.py
```diff
@@ -141,10 +182,56 @@ def test_try_parse_feed_with_beautiful_soup(feed_input: str, is_url: bool, enabl
 
 @pytest.mark.parametrize("enable_cache", [True, False])
 def test_enable_disable_cache(enable_cache):
-    feed_input = "https://rss.nytimes.com/services/xml/rss/nyt/HomePage.xml"
-    rss_controller = RSSController(feed_input=feed_input, enable_cache=enable_cache)
+    controller = RSSController(feed_input=test_urls, enable_cache=enable_cache)
 
     if enable_cache:
-        assert rss_controller._session
+        assert controller._session
```
drobison00, line 188: Unit tests should not check internal class behavior; this breaks concepts of encapsulation and produces brittle tests, because it becomes possible to make changes to the class which are consistent and correct with respect to the API, but still fail.

If we want _session to be part of the public facing API, then it should be explicit. This could be as simple as something like adding a `cache_enabled()` method as well.

tests/controllers/test_rss_controller.py
```diff
@@ -141,10 +182,56 @@ def test_try_parse_feed_with_beautiful_soup(feed_input: str, is_url: bool, enabl
...
+    controller = RSSController(feed_input=feed_input, enable_cache=False)
+
+    with patch("morpheus.controllers.rss_controller.feedparser.parse") as mock_feedparser_parse:
+
+        mock_feedparser_parse.return_value = mock_feed
+
+        with patch.object(controller, '_try_parse_feed') as mock_try_parse_feed:
+            dataframes_generator = controller.parse_feeds()
+            next(dataframes_generator, None)
+            feed_stats: FeedStats = controller._feed_stats_dict[feed_input]
```
drobison00, line 204: Same comment here; we should not be testing against internal class members. This should probably be an interface method to retrieve feed_stats for a given input id.

tests/controllers/test_rss_controller.py
```diff
@@ -141,10 +182,56 @@ def test_try_parse_feed_with_beautiful_soup(feed_input: str, is_url: bool, enabl
...
+            dataframes_generator = controller.parse_feeds()
+            next(dataframes_generator, None)
+
+            feed_stats: FeedStats = controller._feed_stats_dict[feed_input]
+            assert feed_stats.last_try_result == "Failure"
+            assert feed_stats.failure_count == 1
+            assert feed_stats.success_count == 1
+
+            # Resume trying after cooldown period
+            with patch("time.time", return_value=time.time() + controller._cooldown_interval):
```
drobison00, line 229: Same issue here, although is a bit different since we're not testing _cooldown_interval at all. I'd just make this a static value for the test and indicate that it depends on the cooldown interval specified for the stage.

tests/test_rss_source_stage_pipe.py
```diff
@@ -31,12 +31,12 @@
...
 def test_constructor_with_feed_url(config):
 
-    url_feed_input = "https://realpython.com/atom.xml"
+    url_feed_input = "https://fake.nvidia.com/rss/HomePage.xml"
     rss_source_stage = RSSSourceStage(config, feed_input=url_feed_input)
 
     ctlr = rss_source_stage._controller
 
-    assert ctlr._feed_input == {"https://realpython.com/atom.xml"}
+    assert ctlr._feed_input == {"https://fake.nvidia.com/rss/HomePage.xml"}
```
drobison00, line 39: Testing internal class behavior.

tests/test_rss_source_stage_pipe.py
```diff
@@ -59,14 +59,14 @@ def test_constructor_with_feed_file(config):
     assert ctlr._batch_size == 256
     assert rss_source_stage._interval_secs == 5
     assert rss_source_stage._stop_after == 10
-    assert rss_source_stage._max_retries == 2
+    assert rss_source_stage._controller._cooldown_interval == 100
```
drobison00, line 62: Testing internal class behavior.

drobison00 (changes requested):
A couple comments and some change requests related to avoiding testing internal class behavior.

morpheus/controllers/rss_controller.py
```diff
@@ -85,7 +110,7 @@ def _get_response_text(self, url: str) -> str:
         if self._session:
             response = self._session.get(url)
         else:
-            response = requests.get(url, timeout=1.0)
+            response = requests.get(url, timeout=2.0)
```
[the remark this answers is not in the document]
bsuryadevara: Initially tried with 1 sec, but sometimes it's getting timeout while fetching the feed so increased to 2 secs. Ok, I will make it configurable.

tests/controllers/test_rss_controller.py
bsuryadevara: Got it. I will remove private variable access from the tests.

drobison00 (approved):
LGTM

bsuryadevara:
\merge

PR #1296, opened 2023-10-20
Add tests & docstrings for PromptTemplateNode

### LINKED ISSUE #1282: Add documentation and tests for the `PromptTemplateNode`
As part of Sherlock, a prototype of the `PromptTemplateNode` was created. This class will need the following improvements before being merged back to `branch-23.11`:

```[tasklist]
### Tasks
- [x] Add docstrings for the class (including when it should be used and what task it performs)
- [x] Add docstrings to each public method
- [x] Add unit tests for all public functions which execute all code paths
- [x] Add an integration test to verify the component works as part of a pipeline
```

You can find the prototype located [here](https://github.com/nv-morpheus/Morpheus/blob/fea-sherlock/examples/llm/common/template_node.py)

Note: the location above links directly to a commit with the prototype. The current location and functionality of the class may have changed.

dagardner-nv:
## Description
* Move `arvix_source.py` from examples into morpheus itself
* Add CLI registration  and `PreallocatorMixin` for `ArxivSource`
* cache dir and query string are now constructor arguments for `ArxivSource` 
* Add `_utils.llm.execute_node` helper method for executing LLM nodes 
* Expose additional methods for `InputMap` and `LLMContext`
* Add impl for f-string templates in `PromptTemplateNode`
* Add tests for `PromptTemplateNode`

Closes #1282


morpheus/stages/input/arxiv_source.py
```diff
@@ -34,12 +34,24 @@
...
-    def __init__(self, c: Config):
+    """
+    Source stage that downloads PDFs from arxiv and converts them to dataframes
+    
+    Parameters
+    ----------
+    c : `morpheus.config.Config`
+        Pipeline configuration instance.
+    query : `str`, default = "large language models"
+        Query to use for arxiv search.
```
mdemoret-nv, line 47: I dont think there should be a default here. You can use this default in the documentation but it should not be the think we search for automatically

morpheus/stages/input/arxiv_source.py
```diff
@@ -80,10 +92,11 @@ def _generate_frames(self):
         import arxiv
 
         search_results = arxiv.Search(
-            query="large language models",
+            query=self._query,
             max_results=50,
         )
 
+        # TODO: Move this to a config
```
mdemoret-nv, line 99: Lets make this a constructor argument `cache_dir`. If its `None` then no caching is performed. If it is set to a value, then we look up the document and load from the cache. On cache miss, if this is set, we save the document

tests/_utils/llm.py
```diff
@@ -0,0 +1,38 @@
...
+
+    for input_name, input_value in input_values.items():
+        inputs.append(InputMap(f"/{input_name}", input_name))
+        parent_context.set_output(input_name, input_value)
+
+    context = LLMContext(parent_context, "test", inputs)
+
+    asyncio.run(node.execute(context))
+
+    return context.view_outputs
```
mdemoret-nv, line 38: This should probably be: 
```suggestion
    context = parent_context.push("test", inputs)

    context = asyncio.run(node.execute(context))

    return context.view_outputs
```

morpheus/llm/nodes/prompt_template_node.py
```diff
@@ -26,20 +27,24 @@ class PromptTemplateNode(LLMNodeBase):
 
     def __init__(self, template: str, template_format: typing.Literal["f-string", "jinja"] = "f-string") -> None:
```
mdemoret-nv, line 28: Missing documentation on this node and all public members. If sphinx can pull documentation from inherited base classes, we can omit documenting `get_input_names()` and `execute()`

mdemoret-nv (changes requested):

morpheus/stages/input/arxiv_source.py
dagardner-nv: If I understand the code correctly, this isn't a cache dir, as `PyPDFLoader` requires a file path (I don't believe it supports a `StringIO`), 

I think what we could/should do is:
* Make a `download_dir` a constructor argument.
* If `None` call `tempfile.mkdtemp` and on complete call `shutil.rmtree`

dagardner-nv: Alterantely, the DFP stages do:
```
cache_dir: str = "./.cache/dfp"
```

dagardner-nv:
Confirmed methods on subclasses, inherit the parent's docstring in Sphinx if they don't contain their own.

mdemoret-nv (approved):

mdemoret-nv:
@dagardner-nv merge once the Check stage is passing.

PR #1348, opened 2023-11-07
Updated Milvus Tests and Exposed `release_resource` API

### LINKED ISSUE #1345: [BUG]: Milvus 2.3.2 is breaking milvus_vector_db_service tests
### Version

fea-sherlock

Source

### Describe the bug.

milvus==2.3.2 is breaking milvus_vector_db_service tests. Where as it works with `2.3.1`

```
    @pytest.mark.milvus
    def test_delete_by_keys(milvus_service: MilvusVectorDBService,
                            idx_part_collection_config: dict,
                            milvus_data: list[dict]):
        collection_name = "test_delete_by_keys_collection"
    
        # Make sure to drop any existing collection from previous runs.
        milvus_service.drop(collection_name)
    
        # Create a collection.
        milvus_service.create(collection_name, **idx_part_collection_config)
    
        # Insert data into the collection.
        milvus_service.insert(collection_name, milvus_data)
    
        # Delete data by keys from the collection.
        keys_to_delete = [5, 6]
        response = milvus_service.delete_by_keys(collection_name, keys_to_delete)
>       assert response == keys_to_delete
E       assert [] == [5, 6]
E         Right contains 2 more items, first extra item: 5
E         Use -v to get more diff
```

_No response_

### Relevant log output

<details><summary>Click here to see error details</summary><pre>

 [Paste the error here, it will be hidden by default]

</pre></details>

### Full env printout

<details><summary>Click here to see environment details</summary><pre>

 [Paste the results of print_env.sh here, it will be hidden by default]

</pre></details>

### Other/Misc.

_No response_

### Code of Conduct

bsuryadevara:
## Description

- Updated tests to mitigate bug from `milvus==2.3.2`
- Exposed `release_resource` API
- Removed tests that verifies `collection_lock` mechanism

Closes #1345 

### DIFF
--- a/morpheus/service/vdb/milvus_vector_db_service.py
+++ b/morpheus/service/vdb/milvus_vector_db_service.py
@@ -1064,6 +1064,25 @@ class MilvusVectorDBService(VectorDBService):
 
         return resource.describe(**kwargs)
 
+    def release_resource(self, name: str) -> bool:
+        """
+        Release a loaded collection from memory.
+
+        Parameters
+        ----------
+        name : str
+            Name of the collection to release.
+
+        Returns
+        -------
+        bool
+            Returns True if release process is successful, otherwise raises MilvusException.
+        """
+
+        self._client.release_collection(collection_name=name)
+
+        return True
+
     def close(self) -> None:
         """
         Close the connection to the Milvus vector database.
--- a/morpheus/service/vdb/milvus_vector_db_service.py
+++ b/morpheus/service/vdb/milvus_vector_db_service.py
@@ -1064,24 +1064,17 @@ class MilvusVectorDBService(VectorDBService):
 
         return resource.describe(**kwargs)
 
-    def release_resource(self, name: str) -> bool:
+    def release_resource(self, name: str) -> None:
         """
-        Release a loaded collection from memory.
+        Release a loaded collection from the memory.
 
         Parameters
         ----------
         name : str
             Name of the collection to release.
-
-        Returns
-        -------
-        bool
-            Returns True if release process is successful, otherwise raises MilvusException.
         """
 
         self._client.release_collection(collection_name=name)
-
-        return True
 
     def close(self) -> None:
         """
--- a/tests/test_milvus_vector_db_service.py
+++ b/tests/test_milvus_vector_db_service.py
@@ -364,6 +364,24 @@ def test_delete(milvus_service: MilvusVectorDBService, idx_part_collection_confi
     milvus_service.drop(collection_name)
 
 
+def test_release_collection(milvus_service: MilvusVectorDBService,
+                            idx_part_collection_config: dict,
+                            milvus_data: list[dict]):
+    collection_name = "test_release_collection"
+
+    # Make sure to drop any existing collection from previous runs.
+    milvus_service.drop(collection_name)
+
+    # Create a collection.
+    milvus_service.create(collection_name, **idx_part_collection_config)
+
+    # Insert data into the collection.
+    milvus_service.insert(collection_name, milvus_data)
+
+    # Release resource from the memory.
+    milvus_service.release_resource(name=collection_name)
+
+
 def test_get_collection_lock():
     """
     This test doesn't require milvus server to be running.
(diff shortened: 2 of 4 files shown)


morpheus/service/vdb/milvus_vector_db_service.py
```diff
@@ -1064,6 +1064,25 @@ def describe(self, name: str, **kwargs: dict[str, typing.Any]) -> dict:
...
+
+        Returns
+        -------
+        bool
+            Returns True if release process is successful, otherwise raises MilvusException.
+        """
+
+        self._client.release_collection(collection_name=name)
+
+        return True
```
cwharris, line 1084: Question: Why does this return a `bool` if it will always be `True`?

tests/test_milvus_vector_db_service.py
```diff
@@ -322,8 +321,7 @@ def test_delete_by_keys(milvus_service: MilvusVectorDBService,
 
     # Delete data by keys from the collection.
     keys_to_delete = [5, 6]
-    response = milvus_service.delete_by_keys(collection_name, keys_to_delete)
-    assert response == keys_to_delete
+    milvus_service.delete_by_keys(collection_name, keys_to_delete)
```
cwharris, line 324: Question: What changed? What gets returned from `delete_by_keys` now that it no longer returns the deleted keys?

tests/test_milvus_vector_db_service.py
```diff
@@ -366,82 +364,24 @@ def test_delete(milvus_service: MilvusVectorDBService, idx_part_collection_confi
     milvus_service.drop(collection_name)
 
 
-@pytest.mark.milvus
-def test_single_instance_with_collection_lock(milvus_service: MilvusVectorDBService,
```
cwharris, line 370: Question: Were these tests deleted in a previous PR?

cwharris (changes requested):

tests/test_milvus_vector_db_service.py
bsuryadevara: milvus (Embedded server version `2.3.2` for testing) recently released has a bug. It's not returning keys that are being deleted, where as with `milvus==2.3.1` (embedded  & standalone) returns list of keys as a response. So, I have removed assertion  on `delete_by_keys` response and just asserting on the query results after delete operation.

tests/test_milvus_vector_db_service.py
bsuryadevara: It was pushed to `branch-23.11` but not update in the `fea-sherlock`

morpheus/service/vdb/milvus_vector_db_service.py
bsuryadevara: The `release_collection` function from Milvus is not returning any response. Only it throws `MilvusException` if there is any error while releasing the collection from the memory. And also there isn't an API to identify if a collection is loaded/released from the memory. For testing purpose I have added a return statement. 

I have also tried to capture `MilvusException` from the `release_collection` function by passing a collection that does not exists at all, but it is not raising any error.

cwharris: I think it's more appropriate if this function returns nothing. I also think it's fine if we skip the test that tries to reproduce an error case, since we can't reproduce it.

tests/test_milvus_vector_db_service.py
cwharris: Are we sure it's a bug and not just an API change? If so, It may be better to keep use 2.3.1 until there's a fix. If it's just an API change, then the updates make sense to me.

cwharris (changes requested):

tests/test_milvus_vector_db_service.py
bsuryadevara: I'm wondering, If it's just an API, it shouldn't even return the list. And also it has to align with the current version of `pymilvus`. Running delete operation with `pymilvus==2.3.2` and `milvus==2.3.1` returns the same result.

morpheus/service/vdb/milvus_vector_db_service.py
bsuryadevara: Removed return statement.

cwharris (approved):

PR #1382, opened 2023-11-22
Draft: Update conda build dependencies

dagardner-nv:
## Description
* Add missing dependencies to `meta.yaml` and `setup.py`

Closes # 1380

### DIFF
--- a/ci/conda/recipes/morpheus/meta.yaml
+++ b/ci/conda/recipes/morpheus/meta.yaml
@@ -68,6 +68,7 @@ outputs:
       run:
         # Runtime only requirements. This + setup.py is the definitive runtime requirement list
         - {{ pin_compatible('cuda-cudart', min_pin='x.x', max_pin='x') }}
+        - beautifulsoup4=4.12
         - click >=8
         - configargparse 1.5.*
         - cudf
@@ -77,6 +78,8 @@ outputs:
         - dill
         - distributed
         - docker-py 5.0.*
+        - elasticsearch==8.9.0
+        - feedparser=6.0.10
         - grpcio # Version determined from cudf
         - libmrc
         - libwebp>=1.3.2 # Required for CVE mitigation: https://nvd.nist.gov/vuln/detail/CVE-2023-4863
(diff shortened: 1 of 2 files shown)


ci/conda/recipes/morpheus/meta.yaml
```diff
@@ -68,6 +68,7 @@ outputs:
       run:
         # Runtime only requirements. This + setup.py is the definitive runtime requirement list
         - {{ pin_compatible('cuda-cudart', min_pin='x.x', max_pin='x') }}
+        - beautifulsoup4=4.12
```
mdemoret-nv, line 71: I feel like this is only used by a few node types. Should we have an include guard instead of adding it to every install?

ci/conda/recipes/morpheus/meta.yaml
```diff
@@ -77,6 +78,8 @@ outputs:
         - dill
         - distributed
         - docker-py 5.0.*
+        - elasticsearch==8.9.0
+        - feedparser=6.0.10
```
mdemoret-nv, line 82: Same here, not every use will need these dependencies. Should we have an include guard instead of adding it to every install?

mdemoret-nv (commented):
Have some questions. We can add these dependencies but they arent used in many situations. Can we get by with an import guard instead to instruct the user to install these packages if this particular node type is used?

ci/conda/recipes/morpheus/meta.yaml
dagardner-nv: It's used by the `RSSController`, so yes, there is only a single stage in Morpheus proper which requires it. 

However, by this logic the same would apply to `feedparser`

ci/conda/recipes/morpheus/meta.yaml
dagardner-nv: Is your primary concern the dependencies which appear here in the `meta.yaml` as opposed to those in the `cuda11.8_dev.yml` file?

mdemoret-nv (approved):
Looks better now. Will eventually need to re-address some of these in #1389

PR #1390, opened 2023-11-28
Fix race condition in the C++ impl for the pre-process fil stage

dagardner-nv:
## Description
* Fixes a race condition where the `fix_bad_columns` method attempts to fix the same column more than once per-DF, resulting in a failure on the second attempt.


drobison00 (approved):
LGTM, also  fixes the ab_nvsmi failure on my end.

mdemoret-nv (approved):
@dagardner-nv Can you move this out of draft so it can be merged?

PR #1448, opened 2023-12-28
Install headers & morpheus-config.cmake

### LINKED ISSUE #331: [FEA] Allow users to write, compile, and load custom sources/stages/sinks without recompiling all of morpheus

dagardner-nv:
## Description
* Allows users to create C++ stages/messages without a source code checkout of morpheus.
* Explicitly install cpython extensions as a part of the python install.
* Remove usage of `YAPF_EXCLUDE_FLAGS` work-around for an old yapf bug which has been fixed in our current version.
* Update C++ developer guide examples to optionally build independently of morpheus.
* Install versioneer as a conda package, remove in-repo copy of versioneer.py

Closes #331 


cmake/package_search/Finducx.cmake
```diff
@@ -1,3 +1,18 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
mdemoret-nv, line 1: This should include 2023 if the file existed in that year.

cmake/dependencies.cmake
```diff
@@ -46,6 +46,21 @@ if(MORPHEUS_BUILD_BENCHMARKS)
...
+# grpc-repo
+# =========
+rapids_find_package(gRPC REQUIRED
+  GLOBAL_TARGETS
+  gRPC::address_sorting gRPC::gpr gRPC::grpc gRPC::grpc_unsecure gRPC::grpc++ gRPC::grpc++_alts gRPC::grpc++_error_details gRPC::grpc++_reflection
+  gRPC::grpc++_unsecure gRPC::grpc_plugin_support gRPC::grpcpp_channelz gRPC::upb gRPC::grpc_cpp_plugin gRPC::grpc_csharp_plugin gRPC::grpc_node_plugin
+  gRPC::grpc_objective_c_plugin gRPC::grpc_php_plugin gRPC::grpc_python_plugin gRPC::grpc_ruby_plugin
+  BUILD_EXPORT_SET ${PROJECT_NAME}-core-exports
+  INSTALL_EXPORT_SET ${PROJECT_NAME}-core-exports
+)
```
mdemoret-nv, line 62: Make an issue to move this code into the utilities repo

docs/source/developer_guide/guides/4_source_cpp_stage.md
```diff
@@ -511,7 +523,8 @@ Lastly, our `_build_source` method needs to be updated to build a C++ node when
 ```python
 def _build_source(self, builder: mrc.Builder) -> mrc.SegmentObject:
     if self._build_cpp_node():
-        from _lib import morpheus_rabbit as morpheus_rabbit_cpp
+        # pylint: disable=c-extension-no-member,no-name-in-module
```
mdemoret-nv, line 526: I think there is a pylint setting that can help with this issue.

examples/developer_guide/3_simple_cpp_stage/.gitattributes
```diff
@@ -0,0 +1 @@
+src/simple_cpp_stage/_version.py export-subst
```
mdemoret-nv, line 1: Just to be safe, remove this change, and then run `versioneer install --vendor` and see if it re-adds this.

examples/developer_guide/3_simple_cpp_stage/CMakeLists.txt
```diff
@@ -13,52 +13,59 @@
...
 
-pybind11_add_module(morpheus_example MODULE "${CMAKE_CURRENT_SOURCE_DIR}/_lib/pass_thru.cpp")
+# Set the cache to be the same to allow for CCache to be used effectively
+set(MORPHEUS_CACHE_DIR "${CMAKE_SOURCE_DIR}/.cache" CACHE PATH "Directory to contain all CPM and CCache data")
+mark_as_advanced(MORPHEUS_CACHE_DIR)
 
-add_library(${PROJECT_NAME}::morpheus_example ALIAS morpheus_example)
+# Add the Conda environment to the prefix path and add the CMake files
+list(PREPEND CMAKE_PREFIX_PATH "$ENV{CONDA_PREFIX}")
+list(PREPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/../../../external/utilities/cmake")
```
mdemoret-nv, line 26: Lets export any functions needed by the examples from the Morpheus config.cmake file so users dont need to download or know about the necessary utilities functions. They should all be wrapped and exported in the config file.

mdemoret-nv (changes requested):

examples/developer_guide/3_simple_cpp_stage/.gitattributes
dagardner-nv: Confirmed it re-adds it. 
Also the mystery of the `versioneer.py` copy in the root of the repo is solved. 

When you run `versioneer install --vendor` it makes a copy of `versioneer.py` in the repo, allowing the package to not be a runtime or build dependency.

Running `versioneer install --no-vendor` doesn't create the copy.

cmake/dependencies.cmake
```diff
@@ -74,6 +82,10 @@ morpheus_utils_configure_pybind11()
 # =====
 morpheus_utils_configure_rdkafka()
 
+# RxCpp
+# =====
+morpheus_utils_configure_rxcpp()
```
cwharris, line 87: This should not be necessary, as MRC should be pulling in these dependencies automatically. However we have some exports mixed up in MRC due to some recent changes, so the targets are not getting exported correctly. We should revisit the MRC changes and make sure that any publicly linked targets are also exported as dependencies such that Morpheus can pick up on them.

cmake/dependencies.cmake
```diff
@@ -17,44 +17,52 @@ list(APPEND CMAKE_MESSAGE_CONTEXT "dep")
...
 endif()
 
+# glog
+# ====
+morpheus_utils_configure_glog()
+
+# grpc
+# =========
+morpheus_utils_configure_grpc()
+
```
cwharris, line 56: glog and grpc should both be picked up transiently through MRC. we just need to make sure MRC is exporting them correctly.

glog is less important to expose, but we could still do it.

cwharris (changes requested):

cmake/dependencies.cmake
cwharris: We can test this pretty easily using the cyber-dev devcontainer environment, where we can build morpheus and mrc in the same conda environment and have morpheus depend on the locally built mrc.

https://github.com/cwharris/cyber-dev

I have some local changes that need to be merged in to cyber-dev, but if you go this route I'm happy to help you set up. :)

dagardner-nv: This is true, however since Morpheus makes direct use of RxCPP, Glog and gRPC I think we should still be explicitly listing these as dependencies.

cmake/dependencies.cmake
dagardner-nv: Same as above

morpheus/_lib/cmake/libmorpheus.cmake
```diff
@@ -133,6 +143,8 @@ target_sources(morpheus
   PUBLIC
     FILE_SET public_headers
     TYPE HEADERS
+    BASE_DIRS
+  "${CMAKE_CURRENT_SOURCE_DIR}/include"
```
mdemoret-nv, line 147: Put these on the same line

morpheus/_lib/cmake/libmorpheus.cmake
```diff
@@ -165,17 +177,74 @@ set_target_properties(morpheus
...
+    TARGETS morpheus
+    EXPORT
+    ${PROJECT_NAME}-core-exports
+    LIBRARY
+    DESTINATION ${lib_dir}
+    FILE_SET
     public_headers
-  COMPONENT
-    Wheel
+    COMPONENT Core
```
mdemoret-nv, line 201: Formatting was lost here

mdemoret-nv, line 201: Why was the component switched from `Wheel` to `Core`?

morpheus/_lib/cmake/libmorpheus.cmake
```diff
@@ -165,17 +177,74 @@ set_target_properties(morpheus
...
+
+set(rapids_project_version_compat SameMinorVersion)
+
+# Install cmake helper scripts needed to build examples
+set(cmake_lib_dir "${lib_dir}/cmake/morpheus")
+
+install(DIRECTORY
+  "${MORPHEUS_UTILS_ROOT_PATH}"
+  DESTINATION "${cmake_lib_dir}"
+  COMPONENT Core
```
mdemoret-nv, line 219: I dont believe that we need the `Component` sections anymore

mdemoret-nv (changes requested):
Nearly there. Some questions and minor cleanup items

ci/scripts/python_checks.sh
```diff
@@ -59,7 +59,7 @@ if [[ -n "${MORPHEUS_MODIFIED_FILES}" ]]; then
 
    if [[ "${SKIP_YAPF}" == "" ]]; then
       # Run yapf. Will return 1 if there are any diffs
-      YAPF_OUTPUT=`python3 -m yapf --style ${PY_CFG} ${YAPF_EXCLUDE_FLAGS} --diff ${MORPHEUS_MODIFIED_FILES[@]} 2>&1`
+      YAPF_OUTPUT=`python3 -m yapf --style ${PY_CFG} --diff ${MORPHEUS_MODIFIED_FILES[@]} 2>&1`
```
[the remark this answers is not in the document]
dagardner-nv: In the past we had to explicitly specify which files we were excluding from yapf on the command line.  This was a work-around for a yapf bug (https://github.com/google/yapf/issues/984) preventing us from defining these in the `pyproject.toml` file. That bug has been fixed in our current version.

cmake/dependencies.cmake
mdemoret-nv: Since we use these directly, we should have them listed in our Cmake files. If MRC stops exporting them and we didnt explicitly import the libraries, our build would break.

mdemoret-nv: After discussing this further, this seems to be related to a regression where MRC no longer exports the necessary gRPC pieces to allow Morpheus to build. So while we directly use glog and RxCpp, we dont directly use gRPC. So we should tackle the root cause for why gRPC is no longer exported.

docker/Dockerfile
```diff
@@ -187,9 +184,7 @@ RUN --mount=type=cache,id=conda_pkgs,target=/opt/conda/pkgs,sharing=locked \
     conda config --env --set channel_alias ${CONDA_CHANNEL_ALIAS} &&\
     /opt/conda/bin/mamba env update -n morpheus --file docker/conda/environments/cuda${CUDA_MAJOR_VER}.${CUDA_MINOR_VER}_dev.yml &&\
     # Remove channel_alias to use the normal channel in the container
-    conda config --env --remove-key channel_alias &&\
-    # Clean and activate
-    conda clean -afy
```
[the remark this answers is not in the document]
dagardner-nv: Because we perform a conda clean -afy in the development stage, so I thought this was redundant.

cmake/package_search/Finducx.cmake
```diff
@@ -0,0 +1,73 @@
...
+  endforeach()
+
+  if(NOT TARGET ucx::ucx)
+    # Combined ucx::ucx target
+    add_library(ucx::ucx INTERFACE IMPORTED GLOBAL)
+    set_target_properties(ucx::ucx PROPERTIES
+      INTERFACE_LINK_LIBRARIES "${all_ucx_targets}"
+    )
+  endif()
+endif()
```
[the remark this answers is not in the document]
dagardner-nv: I don't know, this was included in the original `mdd_relocatable-package` branch.

cmake/dependencies.cmake
dagardner-nv: I was mistaken, Morpheus doesn't make direct use of gRPC

morpheus/_lib/cmake/libmorpheus.cmake
```diff
@@ -109,6 +108,17 @@ target_include_directories(morpheus
...
+get_target_property(cudf_helpers_include ${cudf_helpers_target} INTERFACE_INCLUDE_DIRECTORIES)
+
+target_include_directories(morpheus
+  PRIVATE
+    ${cudf_helpers_include}
+)
+
+# Also add a dependency to the target so that the headers are generated before the target is built
+add_dependencies(morpheus ${cudf_helpers_target})
+
```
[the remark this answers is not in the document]
dagardner-nv: Not sure, this is also a change inherited from the original `mdd_relocatable-package` branch. 

My guess is that libmorpheus doesn't need to link to the cudf_helpers so because it is a python C-extension which is loaded dynamically by the interpreter when we call `CudfHelper::load()` from our other python modules.

morpheus/_lib/cmake/libmorpheus.cmake
dagardner-nv: I was trying to follow MRC as an example. It was my understanding that the `Wheel` is all the parts of Morpheus that get installed with the python package. 

While `Core` was used for items such as the headers which are installed by cmake with `--target install`

cmake/package_search/Finducx.cmake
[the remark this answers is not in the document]
mdemoret-nv: Can you confirm whether or not this is being used? It really shouldnt be in the Morpheus repo since we do not use UCX directly from Morpheus at all. Might have just been an artifact or debug code.

docker/Dockerfile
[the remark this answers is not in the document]
mdemoret-nv: Fair enough

morpheus/_lib/cmake/libmorpheus.cmake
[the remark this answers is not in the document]
mdemoret-nv: Ah you are correct. I think this was part of some changes I was working on where I was trying to remove the need for the cudf_helpers library entirely. However, I couldnt get that to work 100% but this part stuck.

mdemoret-nv (approved):
Once we confirm whether or not the UCX file is being used, this should be good to go.

cwharris (approved):

PR #1479, opened 2024-01-25
Improve pipeline stop logic to ensure join is called exactly once for all stages

### LINKED ISSUE #1477: [BUG]: `stop()` called twice for each stage on Control-C
### Version

23.07

Conda

### Describe the bug.

When using `Control-C` to stop a pipeline, the stop() method on each stage is being called twice.

1. Set up debug environment for Production DFP pipeline examples:
https://github.com/nv-morpheus/Morpheus/tree/branch-24.03/examples/digital_fingerprinting/production

2. Add breakpoint here:
https://github.com/nv-morpheus/Morpheus/blob/branch-24.03/morpheus/pipeline/stage_base.py#L448

3. Run any of the example pipelines.

4. Send SIGINT (Ctrl-C) to pipeline process.

Observe with breakpoint that stages are looped through twice to call stop().

### Relevant log output

_No response_

### Full env printout

_No response_

### Other/Misc.

_No response_

### Code of Conduct

efajardo-nv:
## Description
1. Removes the `_is_built`, `_is_started` and `_is_stopped` flags and replaces with single member which holds onto the state enum for: INITIALIZED, BUILT, STARTED, STOPPED, COMPLETED
1. Changes the meaning of `stop()` and the meaning of `join()` for stages
   1. `stop()` called 0 or 1 times. Only way it can get called is if `pipeline.stop()` was called indicating the pipeline should try to shut down gracefully. 
      1. Users should only implement this method if they have a source stage (or sources in their stage)
   1. `join()` called exactly 1 time. Only called when the pipeline is complete and all stages are shut down. This is where users should implement any cleanup code
1. Tests for handling all of these scenarios with the pipeline.

Closes #1477


morpheus/pipeline/pipeline.py
```diff
@@ -397,6 +398,13 @@ def stop(self):
         Stops all running stages and the underlying MRC pipeline.
         """
 
+        # Only execute this once
```
drobison00, line 401: Is it possible to have a race here? Do we need a mutex?

efajardo-nv: seems unlikely but possible. i can add a mutex to be safe. do you think i should also add mutex for build and start? cc: @mdemoret-nv

mdemoret-nv (changes requested):
I think your PR makes the fixes specified in the issue but looking at the code in here, I think we can do more to improve how the pipeline state is handled. I have the following recommendations:

1. Convert the `_is_built`, `_is_started` and `_is_stopped` flags into a single member which holds onto the state enum.
   1. An enum works well for the state because the pipeline state should only progress forwards
   1. The enum values should look something like: initialized, built, running, stopping, completed
   1. Changes to the state value should be guarded by a mutex to prevent changing the state from multiple threads
1. We should move all of the logic in the `join()` method into `_start()` after the call to `self._mrc_executor.start()`
   1. This is necessary because there are functions in `join()` which need to be called 100% of the time. However, its not required by the user to call `pipeline.join()`. Its its possible with the current API that stages will never have their `stop()` and `join()` method called.
   1. To fix this, we just need to introduce an [`asyncio.Event`](https://docs.python.org/3/library/asyncio-sync.html#event) into the pipeline
   1. If you follow the example in the docs, we pretty much want to use the same pattern. 
      1. After `self._mrc_executor.start()` is called, create a new task which immediately calls `self._mrc_executor.join_async()`. This will block the task until the pipeline is complete.
      1. After `join_async()` we should have all of the same code which is currently in `Pipeline.join()` to loop over all stages calling `join()`.
      1. Finally, the task should set the pipeline state to Complete and call `set()` on the event object
      1. All that should remain in `Pipeline.join()` is calling `await self._completion_event.wait()` which will block that method from returning until the pipeline finishes.
1. We should change the meaning of `stop()` and the meaning of `join()` for stages
   1. `stop()` should only get called 0 or 1 times. The only way it should get called is if `pipeline.stop()` was called indicating the pipeline should try to shut down gracefully. 
      1. Users should only implement this method if they have a source stage (or sources in their stage)
   1. `join()` should get called exactly 1 time. It should get called when the pipeline is complete and all stages are shut down. This is where users should implement any cleanup code
1. We should store the order that stages were built into a list and use this list when iterating over all the stages in `stop()` and `join()`
   1. This is a small change but will guarantee that stages are stopped and joined in the same order they are built.
1. We should add edge condition tests for handling all of these scenarios with the pipeline.
   1. We have a few checks but more robust tests would be very powerful here.

morpheus/pipeline/pipeline.py
```diff
@@ -425,9 +425,8 @@ async def post_start(executor):
                 with self._mutex:
                     self._state = PipelineState.COMPLETED
 
-                self._completion_event.set()
-
-        asyncio.create_task(post_start(self._mrc_executor))
+        # asyncio.create_task(post_start(self._mrc_executor))
```
mdemoret-nv, line 428: Commented code

morpheus/pipeline/pipeline.py
```diff
@@ -94,7 +94,7 @@ def __init__(self, config: Config):
 
         self._loop: asyncio.AbstractEventLoop = None
 
-        self._completion_event: asyncio.Event = None
+        self._post_start_future: asyncio.Future = None
```
mdemoret-nv, line 97: Add a comment on what this is used for

morpheus/pipeline/pipeline.py
```diff
@@ -452,14 +451,13 @@ async def join(self):
         Suspend execution all currently running stages and the MRC pipeline.
         Typically called after `stop`.
         """
-        await self._completion_event.wait()
+        await self._post_start_future
```
mdemoret-nv, line 454: We need error handling here to ensure the future is set. This should be caught by a test

mdemoret-nv (commented):
Just missing tests on the `join()` method. Do the same tests for both normal and out of order uses.

morpheus/pipeline/pipeline.py
```diff
@@ -448,9 +448,11 @@ def stop(self):
 
     async def join(self):
         """
-        Suspend execution all currently running stages and the MRC pipeline.
-        Typically called after `stop`.
+        Wait until pipeline completes upon which join methods of sources and stages will be called.
         """
+        assert self._post_start_future is not None, "Pipeline must be started before joining"
+        assert self._state != PipelineState.COMPLETED, "Cannot join a pipeline that has already completed"
```
mdemoret-nv, line 454: You can join a pipeline that has completed just find.

tests/pipeline/test_pipeline_state.py
```diff
@@ -130,6 +130,29 @@ async def test_stop_after_stop(config: Config):
...
+
+    pipeline = LinearPipeline(config)
+    assert pipeline.state == PipelineState.INITIALIZED
+    pipeline.set_source(source_test_stage(config))
+    with pytest.raises(Exception) as excinfo:
+        await pipeline.join()
+    assert "must be started" in str(excinfo.value)
+
+
+async def test_join_after_join(config: Config):
```
mdemoret-nv, line 143: Join after join is ok.

tests/pipeline/test_pipeline_state.py
```diff
@@ -242,3 +265,50 @@ async def test_stage_methods_called_stop_after_stop(mock_source_stage_stop,
...
+    mock_source_stage_join.assert_not_called()
+    mock_deserialize_stage_stop.assert_not_called()
+    mock_deserialize_stage_join.assert_not_called()
+
+
+@mock.patch('morpheus.stages.preprocess.deserialize_stage.DeserializeStage.join')
+@mock.patch('morpheus.stages.preprocess.deserialize_stage.DeserializeStage.stop')
+@mock.patch('morpheus.stages.input.in_memory_source_stage.InMemorySourceStage.join')
+@mock.patch('morpheus.stages.input.in_memory_source_stage.InMemorySourceStage.stop')
+async def test_stage_methods_called_join_after_join(mock_source_stage_stop,
```
mdemoret-nv, line 296: join after join is ok.

mdemoret-nv (approved):

PR #1511, opened 2024-02-11
ControlMessage improvements

### LINKED ISSUE #1502: [FEA]: ControlMessage updates supporting TensorMemory references, timestamps, and PyObject caching

New Feature

High

We want Morpheus ControlMessage objects to support additional types of payloads and metadata; specifically, it would be highly useful to be able to directly attach allocated tensor memory structures, and arbitrary PyObject pointers.

This is also a good chance to improve metadata get/set options which are somewhat clunky and hard to iterate.

Proposed pseudo-code

```c++
class ControlMessage
{
  public:
    ...
    set_metadata(std::string key, object) // Make it possible to attach arbitrary objects/data without having to be serializable, and whose lifespan is tied to the ControlMessage
    get_metadata(std::string key, bool fail_if_nonexist);
    
    tensors(std::shared_ptr<TensorMemory>& mem) // Attach tensor memory object
    tensors();  // Retrieve tensor memory

    set_timestamp(std::string key, int timestamp_ns, std::string group = "default"); // store nanosecond timestamps as group::key
    get_timestamp(std::string key, bool fail_if_nonexist);
}
```

### Additional context

_No response_

### Code of Conduct

drobison00:
Resolves #1502 

Adds the ability to attach TensorMemory to ControlMessages
Improves get/set/list metadata functions
Adds the ability to attach grouped/keyed timestamps


morpheus/_lib/include/morpheus/messages/control.hpp
```diff
@@ -271,43 +320,189 @@ class ControlMessage
...
+     *
+     * This method stores a nanosecond precision timestamp associated with a unique
+     * identifier composed of a group and key. If the group::key already exists,
+     * its timestamp will be updated to the new value.
+     *
+     * @param key The specific key for which the timestamp is to be set.
+     * @param group The group to which the key belongs, aiding in namespace separation.
+     * @param timestamp_ns The timestamp value in nanoseconds to be associated with the group::key.
+     */
+    void set_timestamp(const std::string& group, const std::string& key, std::size_t timestamp_ns);
```
mdemoret-nv, line 334: Add overloads that dont require a group and allow users to do the concatenation themselves or to look up without any group.

morpheus/_lib/include/morpheus/messages/control.hpp
```diff
@@ -271,43 +320,189 @@ class ControlMessage
...
     static std::map<std::string, ControlMessageType> s_task_type_map;  // NOLINT
 
     ControlMessageType m_cm_type{ControlMessageType::NONE};
     std::shared_ptr<MessageMeta> m_payload{nullptr};
+    std::shared_ptr<TensorMemory> m_tensors{nullptr};
 
     nlohmann::json m_tasks{};
     nlohmann::json m_config{};
+
+    std::map<std::string, std::size_t> m_timestamps{};
```
mdemoret-nv, line 376: Lets use `std::chrono` timepoints here that are backed by `size_t`. Will improve the readability without adding overhead

morpheus/_lib/messages/module.cpp
```diff
@@ -379,13 +378,50 @@ PYBIND11_MODULE(messages, _module)
              py::arg("config"))
         .def("config", pybind11::overload_cast<ControlMessage&>(&ControlMessageProxy::config))
         .def("copy", &ControlMessageProxy::copy)
-        .def("get_metadata", &ControlMessageProxy::get_metadata, py::arg("key") = py::none())
+        .def(
+            "get_metadata",
+            [](ControlMessage& self, const py::object& key = py::none(), const py::object& default_value = py::none()) {
```
mdemoret-nv, line 383: Dont specify the default values for the lambda function. Only use the `py::arg() = default` syntax

morpheus/_lib/messages/module.cpp
```diff
@@ -379,13 +378,50 @@ PYBIND11_MODULE(messages, _module)
...
+                if (key.is_none())
+                {
+                    // Directly return all metadata
+                    return ControlMessageProxy::get_metadata(self, key, default_value);
+                }
+                else
+                {
+                    // Call the original get_metadata with the provided key and default value
+                    return ControlMessageProxy::get_metadata(self, key.cast<py::str>(), default_value);
+                }
```
mdemoret-nv, line 393: This should be movable to the proxy class

morpheus/_lib/messages/module.cpp
```diff
@@ -379,13 +378,50 @@ PYBIND11_MODULE(messages, _module)
...
+             "Set a timestamp for a given key and group.",
+             py::arg("group"),
+             py::arg("key"),
+             py::arg("timestamp_ns"))
         .def("has_metadata", &ControlMessage::has_metadata, py::arg("key"))
         .def("has_task", &ControlMessage::has_task, py::arg("task_type"))
         .def("list_metadata", &ControlMessageProxy::list_metadata)
         .def("payload", pybind11::overload_cast<>(&ControlMessage::payload), py::return_value_policy::move)
         .def("payload", pybind11::overload_cast<const std::shared_ptr<MessageMeta>&>(&ControlMessage::payload))
+        .def("tensors", pybind11::overload_cast<>(&ControlMessage::tensors), py::return_value_policy::move)
```
mdemoret-nv, line 423: The return value policy needs to reference the parent object. I believe its called `reference_internal`

morpheus/_lib/src/messages/control.cpp
```diff
@@ -41,12 +42,30 @@ ControlMessage::ControlMessage(const nlohmann::json& _config) :
   m_tasks({})
 {
     config(_config);
+    // Register cleanup for this instance
+    auto cleanup = [this]() {
+        py::gil_scoped_acquire acquire;  // Acquire the GIL
+        this->m_tensors.reset();         // Clear the shared_ptr
+    };
```
mdemoret-nv, line 49: We want to be handling this a different way and not with `atexit`

mdemoret-nv (changes requested):

morpheus/_lib/src/messages/control.cpp
drobison00: Yeah, this is removed. It wasn't related to the behavior I was observing.

morpheus/_lib/messages/module.cpp
drobison00: So, reference_internal ties the lifetime of the returned object to the parent. I'm not sure this is the correct behavior.

If we create the tensor_memory, then pass it to the CM, then get it back in some other context, its possible we could have 3 references to the tensor_memory, and any of them should be able to safely go out of scope.

If we create the tensor_memory SP in c++, and then retrieve it in python at a later time, we should still be fine regardless of the order in which they go out of scope.

morpheus/_lib/include/morpheus/messages/control.hpp
```diff
@@ -17,15 +17,19 @@
 
 #pragma once
 
+#include "morpheus/messages/memory/tensor_memory.hpp"
```
mdemoret-nv, line 20: Change this to a forward declare

morpheus/_lib/include/morpheus/messages/control.hpp
```diff
@@ -269,49 +325,185 @@ class ControlMessage
...
+     *
+     * This method stores a nanosecond precision timestamp associated with a unique
+     * identifier composed of a group and key. If the group::key already exists,
+     * its timestamp will be updated to the new value.
+     *
+     * @param key The specific key for which the timestamp is to be set.
+     * @param group The group to which the key belongs, aiding in namespace separation.
+     * @param timestamp_ns The timestamp value in nanoseconds to be associated with the group::key.
+     */
+    void set_timestamp(const std::string& key, std::chrono::nanoseconds timestamp_ns);
```
mdemoret-nv, line 339: Make a typedef for a time_point and use `time_point` instead of `nanoseconds`

mdemoret-nv (approved):

PR #1529, opened 2024-02-21
Upgrade `openai` version to 1.13 and `langchain` to version 0.1.9

mdemoret-nv:
## Description
This PRs main focus is to update the version of the OpenAI package to 1.13 to allow using the built in retry logic and improved code. In addition, the following changes are also made:

- Switch to using `langchain==0.1.9` installed via pip to get around pandas dependency issue
- Sorted the lists in `dependencies.yaml` to make merging easier in the future
- Improved logging for the `OpenAILlmService` to log all requests ordered to a file
- New feature to add default model arguments to the service to avoid repeating arguments for each client. You can now just overwrite the default args
- New feature to allow the `run_ci_local.sh` to persist any changes to the code back to the host via a volume mount.


morpheus/llm/services/openai_chat_service.py
```diff
@@ -72,27 +110,46 @@ def get_input_names(self) -> list[str]:
...
+        duration_ms = (end_time - start_time) * 1000.0
+
+        log_str = dedent("""
+        ============= MESSAGE %d START ==============
+                        --- Input ---
+        %s
+                        --- Output --- (%f ms)
+        %s
+        =============  MESSAGE %d END ==============
+        """).strip("\n")
```
dagardner-nv, line 133: I would be tempted to make this a class property or a global since its a constant, and would avoid the `dedent` and the strip for each call.

morpheus/stages/inference/inference_stage.py
```diff
@@ -443,4 +443,4 @@ def _convert_one_response(output: MultiResponseMessage, inf: MultiInferenceMessa
             for i, idx in enumerate(mess_ids):
                 probs[idx, :] = cp.maximum(probs[idx, :], resp_probs[i, :])
 
-        return MultiResponseMessage.from_message(inf, memory=memory, offset=inf.offset, count=inf.mess_count)
+        return MultiResponseMessage.from_message(inf, memory=memory, offset=seq_offset, count=seq_count)
```
dagardner-nv, line 446: Is this an unrelated bug fix?
Add to the PR description.

tests/conftest.py
```diff
@@ -1046,26 +1045,3 @@ def openai_fixture(fail_missing: bool):
...
-
-@pytest.mark.usefixtures("nemollm")
-@pytest.fixture(name="mock_nemollm")
-def mock_nemollm_fixture():
-    with mock.patch("nemollm.NemoLLM") as mock_nemollm:
-        mock_nemollm.return_value = mock_nemollm
-        mock_nemollm.generate_multiple.return_value = ["test_output"]
-        mock_nemollm.post_process_generate_response.return_value = {"text": "test_output"}
-
-        yield mock_nemollm
```
dagardner-nv, line 1071: Why are these being removed?

dagardner-nv (approved):
Looks good, just had a few questions

morpheus/stages/inference/inference_stage.py
mdemoret-nv: Yes, it never gets hit and I just wanted to fix it.

tests/conftest.py
mdemoret-nv: Reverted this. They are duplicated but for some reason they are slightly different

morpheus/llm/services/openai_chat_service.py
mdemoret-nv: Good point

PR #1539, opened 2024-02-29
Catch langchain agent errors

dagardner-nv:
## Description
* Catch any uncaught/un-handled langchain agent errors, return an error string so that we don't end up with a missing result.
* Fix version conflicts for nemollm and dgl
* Expand existing tests for the `LangChainAgentNode`

### DIFF
--- a/morpheus/llm/nodes/langchain_agent_node.py
+++ b/morpheus/llm/nodes/langchain_agent_node.py
@@ -66,7 +66,12 @@ class LangChainAgentNode(LLMNodeBase):
             return results
 
         # We are not dealing with a list, so run single
-        return await self._agent_executor.arun(**kwargs)
+        try:
+            return await self._agent_executor.arun(**kwargs)
+        except Exception as e:
+            error_msg = f"Error running agent: {e}"
+            logger.exception(error_msg)
+            return error_msg
 
     async def execute(self, context: LLMContext) -> LLMContext:
 
--- a/morpheus/llm/nodes/langchain_agent_node.py
+++ b/morpheus/llm/nodes/langchain_agent_node.py
@@ -58,7 +58,7 @@ class LangChainAgentNode(LLMNodeBase):
             # Run multiple again
             results_async = [self._run_single(**x) for x in input_list]
 
-            results = await asyncio.gather(*results_async, return_exceptions=True)
+            results = await asyncio.gather(*results_async)
 
             # # Transform from list[dict[str, Any]] to dict[str, list[Any]]
             # results = {k: [x[k] for x in results] for k in results[0]}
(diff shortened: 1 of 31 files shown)


morpheus/llm/nodes/langchain_agent_node.py
```diff
@@ -66,9 +66,14 @@ async def _run_single(self, **kwargs: dict[str, typing.Any]) -> dict[str, typing
...
         # We are not dealing with a list, so run single
-        return await self._agent_executor.arun(**kwargs)
-
-    async def execute(self, context: LLMContext) -> LLMContext:
+        try:
+            return await self._agent_executor.arun(**kwargs)
+        except Exception as e:
+            error_msg = f"Error running agent: {e}"
+            logger.exception(error_msg)
+            return error_msg
```
cwharris, line 74: This is fine as is, but what I'd really like to see is the exception caught around `_run_single`, and the ability to set an error on the `LLMContext`, rather than just setting the value to a string with the contents "Error...".

cwharris (approved):

morpheus/llm/nodes/langchain_agent_node.py
cwharris: I'm also a little confused why throwing an exception here would result in a missed value instead of blowing up the entire operation. Is there something upstream catching these errors and ignoring them?

morpheus/llm/nodes/langchain_agent_node.py
```diff
@@ -66,9 +66,14 @@ async def _run_single(self, **kwargs: dict[str, typing.Any]) -> dict[str, typing
             return results
 
         # We are not dealing with a list, so run single
-        return await self._agent_executor.arun(**kwargs)
-
-    async def execute(self, context: LLMContext) -> LLMContext:
+        try:
+            return await self._agent_executor.arun(**kwargs)
```
mdemoret-nv, line 70: Can we try using `asyncio.gather(*results_async)` which returns the exceptions by value instead of throwing an error? And see if that works better.

morpheus/llm/nodes/langchain_agent_node.py
dagardner-nv: > I'm also a little confused why throwing an exception here would result in a missed value instead of blowing up the entire operation. Is there something upstream catching these errors and ignoring them?

Sorry the PR description is a bit mis-leading. You are correct if an exception was raised it would halt the pipeline.

The original version of this PR just did a catch and log without returning a value causing `None` to be stored as the output.

morpheus/llm/nodes/langchain_agent_node.py
dagardner-nv: The issue is that the results are ultimately stored in the context as:
```
context.set_output(results)
```

Which assumes that everything in `results` is implicitly castable to `nlohmann::json`. 
ex:
```py
from morpheus.llm import LLMContext
ctx = LLMContext()
ctx.set_output(['a', RuntimeError('test'), 'b'])
print(ctx.view_outputs)
```

yields:
```
['a', None, 'b']
```

Off the top of my head I'm thinking:
1) Update the outputs to be a `py::object` rather than `nlohmann::json`
2) Go back to just log/return an error message
3) Serialize the exception to a string in the pybind mapping for `set_output`

dagardner-nv: Opened up issues for the long-term fix for this :
* nv-morpheus/MRC#450
* #1560

mdemoret-nv (approved):

PR #1573, opened 2024-03-21
Support `ControlMessage` for `PreProcessNLPStage` `PreProcessFILStage` `AddScoreStageBase`

yuchenz427:
## Description

Support `ControlMessage` as an input type for `PreProcessNLPStage` `PreProcessFILStage` `AddScoreStageBase`


morpheus/_lib/include/morpheus/messages/memory/inference_memory_fil.hpp
```diff
@@ -34,7 +34,7 @@ namespace morpheus {
  * @{
  * @file
  */
-
+#pragma GCC visibility push(default)
```
mdemoret-nv, line 37: You now have 2 pushes and only one pop.

You can also use the macro `MORPHEUS_EXPORT` to declare it visible. Like `class MORPHEUS_EXPORT LLMContext`. I havent had good success with the GCC visibility push/pop and sometimes prefer the macro.

morpheus/_lib/include/morpheus/stages/add_scores_stage_base.hpp
```diff
@@ -68,6 +73,8 @@ class AddScoresStageBase
      */
     AddScoresStageBase(std::map<std::size_t, std::string> idx2label, std::optional<float> threshold);
 
+    std::shared_ptr<OutputT> pre_process_batch(std::shared_ptr<InputT> x);
+
```
mdemoret-nv, line 77: This class already as the `on_data` method. We dont need to add a new method here. Just make `on_data` public so it can be used in tests

morpheus/_lib/include/morpheus/stages/add_scores_stage_base.hpp
```diff
@@ -81,6 +88,181 @@ class AddScoresStageBase
...
+    // sink_type_t not supported
+    else
+    {
+        std::string error_msg{"AddScoresStageBase receives unsupported input type: " + std::string(typeid(x).name())};
+        LOG(ERROR) << error_msg;
+        throw std::runtime_error(error_msg);
+    }
+    return x;
+}
+
```
mdemoret-nv, line 265: This function is getting large. Lets break it up into separate private methods for the two primary code paths.

morpheus/_lib/include/morpheus/stages/add_scores_stage_base.hpp
```diff
@@ -81,6 +88,181 @@ class AddScoresStageBase
...
+                                                        std::optional<float> threshold) :
+  base_t(),
+  m_idx2label(std::move(idx2label)),
+  m_threshold(threshold),
+  m_min_col_count(m_idx2label.rbegin()->first)  // Ordered map's largest key will be the last entry
+{
+    this->pipe(rxcpp::operators::map([this](sink_type_t x) {
+        return this->on_data(std::move(x));
+    }));
+}
```
mdemoret-nv, line 102: Lets switch to using [explicit template instantiation](https://stackoverflow.com/a/2351622) since there are only two supported types for this class. To do this:

1. Move all implementation details back to the cpp file.
2. Add using declarations for the supported types, i.e. `using AddScoresStageBaseMM = AddScoresStageBase<MultiResponseMessage, MultiResponseMessage>;` (You can use `// NOLINT(readability-identifier-naming)` to avoid the linter error)
3. At the end of the cpp file, explicitly instantiate the classes we want: `template class AddScoresStageBase<MultiResponseMessage, MultiResponseMessage>;`

Do this for all of the new templated classes

morpheus/_lib/include/morpheus/stages/add_scores_stage_base.hpp
```diff
@@ -81,6 +88,181 @@ class AddScoresStageBase
...
+        {
+            columns.push_back(column_name);
+            tensors.emplace_back(output_tensor.slice({0, static_cast<TensorIndex>(column_num)},
+                                                     {num_rows, static_cast<TensorIndex>(column_num + 1)}));
+
+            ++i;
+        }
+
+        // A copy of MultiMessage::set_meta(const std::vector<std::string>& column_names, const
+        // std::vector<TensorObject>& tensors)
```
mdemoret-nv, line 207: This should really be moved into a utility function on ControlMessage.

morpheus/_lib/include/morpheus/stages/preprocess_nlp.hpp
```diff
@@ -89,6 +101,8 @@ class PreprocessNLPStage
                        int stride         = -1,
                        std::string column = "data");
 
+    std::shared_ptr<OutputT> pre_process_batch(std::shared_ptr<InputT> x);
```
mdemoret-nv, line 104: Looking at the other classes, lets standardize on `on_data` for the name of the function for each stage which handles messages.

morpheus/_lib/include/morpheus/stages/preprocess_nlp.hpp
```diff
@@ -104,14 +118,230 @@ class PreprocessNLPStage
...
+
+                output.on_next(std::move(next));
+            },
+            [&](std::exception_ptr error_ptr) {
+                output.on_error(error_ptr);
+            },
+            [&]() {
+                output.on_completed();
+            }));
+    };
```
mdemoret-nv, line 171: Lets follow how `AddScoresBase` sets up their subscription and just use the `rxcpp::operators::map` function which is equivalent to what is happening here.

morpheus/_lib/include/morpheus/stages/preprocess_fil.hpp
```diff
@@ -76,23 +88,234 @@ class PreprocessFILStage
...
+            [&output, this](sink_type_t x) {
+                auto next = this->pre_process_batch(x);
+                output.on_next(std::move(next));
+            },
+            [&](std::exception_ptr error_ptr) {
+                output.on_error(error_ptr);
+            },
+            [&]() {
+                output.on_completed();
+            }));
```
mdemoret-nv, line 111: Switch to using `rxcpp::operators::map`

morpheus/_lib/include/morpheus/stages/preprocess_fil.hpp
```diff
@@ -76,23 +88,234 @@ class PreprocessFILStage
...
+    };
+}  // namespace morpheus
+
+void transform_bad_columns(std::vector<std::string>& fea_cols,
+                           morpheus::MutableTableInfo& mutable_info,
+                           std::vector<std::string>& df_meta_col_names);
```
mdemoret-nv, line 117: What is this function for?

- If its a public global namespaced function, it should be in one of the utils files.
- If its tied to this particular class, make it a static function.
- And if it is a public function, it needs to have doxygen comments.

morpheus/_lib/include/morpheus/stages/preprocess_fil.hpp
```diff
@@ -76,23 +88,234 @@ class PreprocessFILStage
...
+        return next;
+    }
+    // sink_type_t not supported
+    else
+    {
+        std::string error_msg{"PreProcessFILStage receives unsupported input type: " + std::string(typeid(x).name())};
+        LOG(ERROR) << error_msg;
+        throw std::runtime_error(error_msg);
+    }
+}
```
mdemoret-nv, line 289: Break these long functions into 2 smaller implementations

morpheus/_lib/tests/stages/test_add_scores.cpp
```diff
@@ -0,0 +1,66 @@
...
+    cm->tensors(cm_tensor_memory);
+
+    // Create PreProcessControlMessageStage
+    auto cm_stage = std::make_shared<AddScoresStage<ControlMessage, ControlMessage>>(idx2label);
+
+    auto cm_response              = cm_stage->pre_process_batch(cm);
+    auto cm_response_probs_tensor = cm_response->tensors()->get_tensor("probs");
+
+    // Check the returned tensors have the same size
+    EXPECT_EQ(multi_response_probs_tensor.count(), cm_response_probs_tensor.count());
```
mdemoret-nv, line 65: We need to verify more than just the count is the same. Lets compare the contents, and verify against a known correct response.

morpheus/_lib/tests/stages/test_add_classification.cpp
```diff
@@ -0,0 +1,75 @@
...
+    cm->tensors(cm_tensor_memory);
+
+    // Create PreProcessControlMessageStage
+    auto cm_stage = std::make_shared<AddClassificationsStage<ControlMessage, ControlMessage>>(idx2label, 0.0);
+
+    auto cm_response              = cm_stage->pre_process_batch(cm);
+    auto cm_response_probs_tensor = cm_response->tensors()->get_tensor("probs");
+
+    // Check the returned tensors have the same size
+    EXPECT_EQ(multi_response_probs_tensor.count(), cm_response_probs_tensor.count());
```
mdemoret-nv, line 74: We need to verify more than just the count is the same. Lets compare the contents, and verify against a known correct response.

morpheus/_lib/tests/stages/test_preprocess_nlp.cpp
```diff
@@ -0,0 +1,77 @@
...
+                                                                                  false /*truncation*/,
+                                                                                  false /*do_lower_case*/,
+                                                                                  false /*add_special_token*/,
+                                                                                  1 /*stride*/,
+                                                                                  "country" /*column*/);
+    auto multi_response         = multi_stage->pre_process_batch(multi);
+    auto multi_response_payload = multi_response->meta;
+
+    // Check if identical number of rows are returned
+    EXPECT_EQ(multi_response_payload->count(), cm_response_payload->count());
```
mdemoret-nv, line 76: We need to verify more than just the count is the same. Lets compare the contents, and verify against a known correct response.

morpheus/pipeline/single_port_stage.py
```diff
@@ -64,6 +66,14 @@ def _pre_compute_schema(self, schema: _pipeline.StageSchema):
...
             raise RuntimeError((f"The {self.name} stage cannot handle input of {input_type}. "
                                 f"Accepted input types: {self.accepted_types()}"))
+        out_type = MultiResponseMessage
+        if (schema.input_type == ControlMessage):
+            self._use_control_message = True
+            out_type = ControlMessage
+        else:
+            self._use_control_message = False
+
+        schema.output_schema.set_type(out_type)
```
mdemoret-nv, line 76: This is a bit too general but I like the direction. Not all return types for every class that derives from `SinglePortStage` will return a `MultiResponseMessage` or a `ControlMessage`. But we could save off the input/output type to make it easier to use in derived classes. Something like this would work:

```suggestion
        self._input_type = schema.input_type
```

And then you can just use `if (self._input_type == ControlMessage)` in derived classes

mdemoret-nv (changes requested):
A great start. Just a lot of little things that need to be cleaned up. Outside of the comments, you are missing python tests for the add scores and add classification stages that use `ControlMessage`.

morpheus/_lib/include/morpheus/stages/preprocess_fil.hpp
yuchenz427: I abstracted this out from the original `fix_bad_columns()` to avoid duplicate code. Do you have any specific reason that this should be a static function rather than non-static?

morpheus/_lib/tests/stages/test_preprocess_fil.cpp
```diff
@@ -19,35 +36,51 @@ TEST_F(TestPreprocessFIL, TestProcessControlMessageAndMultiMessage)
...
-    // Check if identical number of rows are returned
-    EXPECT_EQ(multi_response_payload->count(), cm_response_payload->count());
+    // Col1 in MatxUtil__MatxCreateSegIds is not initialized
+    // auto cm_seq_ids = cm_tensors->get_tensor("seq_ids");
+    // auto mm_seq_ids = mm_tensors->get_tensor("seq_ids");
+    // std::vector<TensorIndex> cm_seq_ids_host(cm_seq_ids.count());
+    // std::vector<TensorIndex> mm_seq_ids_host(mm_seq_ids.count());
+    // MRC_CHECK_CUDA(cudaMemcpy(cm_seq_ids_host.data(), cm_seq_ids.data(), cm_seq_ids.count() * sizeof(TensorIndex),
+    // cudaMemcpyDeviceToHost)); MRC_CHECK_CUDA(cudaMemcpy(mm_seq_ids_host.data(), mm_seq_ids.data(), mm_seq_ids.count()
+    // * sizeof(TensorIndex), cudaMemcpyDeviceToHost)); EXPECT_EQ(cm_seq_ids_host, mm_seq_ids_host);
```
mdemoret-nv, line 85: To enable this check to function, update `MatxUtil__MatxCreateSegIds` to initialize column 1 to 0 so we can consistently check the output without uninitialized values.

mdemoret-nv (approved):
Once the tests are finished and the comments are resolved, we can merge.

PR #1637, opened 2024-04-17
Update `ControlMessage` to hold arbitrary Python objects & update `MessageMeta` to copy & slice

yuchenz427:
## Description

- For ControlMessage, use a specialization of `nlohmann::basic_json` to hold arbitrary Python objects for `m_metadata` & `m_tasks`

- Implement dataframe slicing method for `MessageMeta` which is equivalent to copy_meta_ranges() in MultiMessage. The method will do copy & slicing to the original dataframe, instead of sharing the ownership.


morpheus/_lib/include/morpheus/messages/control.hpp
```diff
@@ -374,6 +389,7 @@ class ControlMessage
 
     nlohmann::json m_tasks{};
     nlohmann::json m_config{};
```
mdemoret-nv, line 391: Dont we want to be removing the JSON objects from this class and replacing them with `JSONValues`?

morpheus/_lib/messages/module.cpp
```diff
@@ -427,7 +429,9 @@ PYBIND11_MODULE(messages, _module)
         .def("set_metadata", &ControlMessageProxy::set_metadata, py::arg("key"), py::arg("value"))
         .def("task_type", pybind11::overload_cast<>(&ControlMessage::task_type))
         .def(
-            "task_type", pybind11::overload_cast<ControlMessageType>(&ControlMessage::task_type), py::arg("task_type"));
+            "task_type", pybind11::overload_cast<ControlMessageType>(&ControlMessage::task_type), py::arg("task_type"))
+        .def("set_py_object", &ControlMessage::set_py_object, py::arg("path"), py::arg("value"))
+        .def("get_py_object", &ControlMessage::get_py_object, py::arg("path"));
```
mdemoret-nv, line 434: No need for binding the functions `get_py_object` and `set_py_object`. They only need to exist to allow C++ classes to pull python objects from C++. This would happen automatically from Python

mdemoret-nv (changes requested):
- `set_metadata` and `get_metadata` should return a `JSONValues` object.
- Same with Tasks

morpheus/_lib/include/morpheus/messages/control.hpp
```diff
@@ -166,6 +166,37 @@ class TensorMemory;
...
+    PythonByteContainer() = default;
+    PythonByteContainer(pybind11::object py_obj) : m_py_obj(std::move(py_obj)) {}
+
+    pybind11::object get_py_obj() const
+    {
+        return m_py_obj;
+    }
+
+  private:
+    pybind11::object m_py_obj;
```
mdemoret-nv, line 181: We will want to change `pybind11::object` here to the `PyHolder` class in `pymrc`. `PyHolder` wraps a `pybind11::object` and makes sure that we have the GIL when its destroyed (which would throw an error otherwise`.

morpheus/_lib/src/messages/control.cpp
```diff
@@ -34,18 +34,146 @@
...
+        return std::move(dict);
+    }
+    if (source.is_string())
+    {
+        return py::str(source.get<std::string>());
+    }
+    if (source.is_binary())
+    {
+        return source.get_binary().get_py_obj();
+    }
```
mdemoret-nv, line 86: To be safe, we should utilize the type ID property of the binary objects to ensure that our binary type is definitely a python one. For example,

```cpp
nlohmann::json j = {{"key", "value"}};

j["test"] = nlohmann::json::binary(morpheus::PythonByteContainer(source), typeid(py::object));

if (j["test"].is_binary())
{
    if (j["test"].get_binary().has_subtype() && j["test"].get_binary().subtype() == typeid(py::object))
    {
        return j["test"].get_binary().get_py_obj();
    }

    // Otherwise, throw an error
}
```

tests/messages/test_control_message.py
```diff
@@ -400,3 +400,26 @@ def test_consistency_after_multiple_operations():
...
+    message = messages.ControlMessage()
+
+    non_serializable_obj = NonSerializablePyObj()
+    message.set_metadata("non_serializable_py_obj", non_serializable_obj)
+    assert message.get_metadata("non_serializable_py_obj") is non_serializable_obj
+    
+    message.add_task("non_serializable", {"non_serializable_task": non_serializable_obj})
+    assert message.has_task("non_serializable")
+    assert message.get_tasks()["non_serializable"][0]["non_serializable_task"] is non_serializable_obj
+    assert message.remove_task("non_serializable")["non_serializable_task"] is non_serializable_obj
```
mdemoret-nv, line 425: This is great but lets expand on this. We updated both `tasks` and `metadata` to use the new json type. We should add tests for both.

mdemoret-nv (commented):
Looking good but lets build on this. Other things we should be thinking of here:
- How could we possibly break this code? Lets make tests to try and break it and fix the implementation if they succeed. Try different types of objects, overwriting objects, deleting objects, etc.
- Can we verify that the python object was destroyed if we erase the key that is holding it?
- What other utility functions would developers want when working with these objects?

morpheus/_lib/include/morpheus/messages/control.hpp
```diff
@@ -170,15 +171,15 @@ class PythonByteContainer : public std::vector<uint8_t>
...
   public:
     PythonByteContainer() = default;
-    PythonByteContainer(pybind11::object py_obj) : m_py_obj(std::move(py_obj)) {}
+    PythonByteContainer(mrc::pymrc::PyHolder py_obj) : m_py_obj(std::move(py_obj)) {}
 
-    pybind11::object get_py_obj() const
+    mrc::pymrc::PyHolder get_py_obj() const
     {
         return m_py_obj;
     }
```
mdemoret-nv, line 179: Move the implementation of these functions into a .cpp file

morpheus/_lib/include/morpheus/messages/control.hpp
```diff
@@ -170,15 +171,15 @@ class PythonByteContainer : public std::vector<uint8_t>
 {
   public:
     PythonByteContainer() = default;
-    PythonByteContainer(pybind11::object py_obj) : m_py_obj(std::move(py_obj)) {}
+    PythonByteContainer(mrc::pymrc::PyHolder py_obj) : m_py_obj(std::move(py_obj)) {}
```
mdemoret-nv, line 174: Move the `PythonByteContainer` class and `json_t` using declaration to its own file.

mdemoret-nv (changes requested):
Update the type casters in `include/morpheus/pybind11/json.hpp` to use your new type.

morpheus/_lib/include/morpheus/utilities/json_types.hpp
```diff
@@ -17,7 +17,63 @@
...
+MORPHEUS_EXPORT py::object cast_from_json(const morpheus::utilities::json_t& source);
+MORPHEUS_EXPORT json_t cast_from_pyobject(const py::object& source, mrc::pymrc::unserializable_handler_fn_t unserializable_handler_fn);
+MORPHEUS_EXPORT json_t cast_from_pyobject(const py::object& source);
+
+// NOLINTBEGIN(readability-identifier-naming)
+/*
+    Derived class from json_t to allow for custom type names. Use this if the return type would always be an object
+   (i.e. dict[str, Any] in python)
+*/
+class MORPHEUS_EXPORT json_t_dict : public morpheus::utilities::json_t
```
mdemoret-nv, line 66: this should be `json_dict_t`

morpheus/_lib/include/morpheus/utilities/json_types.hpp
```diff
@@ -17,7 +17,63 @@
...
+{};
+
+/*
+    Derived class from json_t to allow for custom type names. Use this if the return type would always be an object
+   (i.e. dict[str, Any] in python)
+*/
+class MORPHEUS_EXPORT json_t_list : public morpheus::utilities::json_t
```
mdemoret-nv, line 73: `json_list_t`

mdemoret-nv (approved):

PR #1695, opened 2024-05-10
Handling agent parsing error and adding a default value for NVAI_BASE_URL

hsin-c:
The agent model in the CVE pipeline sometimes runs into a parsing error that crashes the entire run. This PR adds the handling of such error by replacing it with a default failure sentence and logging the failure.
This PR also includes an addition of a default value for `NVAI_BASE_URL`.

### DIFF
--- a/morpheus/llm/nodes/langchain_agent_node.py
+++ b/morpheus/llm/nodes/langchain_agent_node.py
@@ -37,7 +37,10 @@ class LangChainAgentNode(LLMNodeBase):
         The agent executor to use to execute.
     """
 
-    def __init__(self, agent_executor: "AgentExecutor", replace_exceptions: bool=False, replace_exceptions_value: typing.Optional[str]=None):
+    def __init__(self,
+                 agent_executor: "AgentExecutor",
+                 replace_exceptions: bool = False,
+                 replace_exceptions_value: typing.Optional[str] = None):
         super().__init__()
 
         self._agent_executor = agent_executor
@@ -94,10 +97,8 @@ class LangChainAgentNode(LLMNodeBase):
                         # If the agent encounters a parsing error or a server error after retries, replace the error
                         # with a default value to prevent the pipeline from crashing
                         results[i][j] = self._replace_exceptions_value
-                        logger.warning(
-                            f"Exception encountered in result[{i}][{j}]: {answer}. "
-                            f"Replacing with default message: \"{self._replace_exceptions_value}\"."
-                        )
+                        logger.warning(f"Exception encountered in result[{i}][{j}]: {answer}. "
+                                       f"Replacing with default message: \"{self._replace_exceptions_value}\".")
 
         context.set_output(results)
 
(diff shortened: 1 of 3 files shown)


morpheus/llm/nodes/langchain_agent_node.py
```diff
@@ -78,6 +78,17 @@ async def execute(self, context: LLMContext) -> LLMContext:  # pylint: disable=i
...
+        default_message = "I do not have a definitive answer for this checklist item."
+        for i, answer_list in enumerate(results):
+            for j, answer in enumerate(answer_list):
+                if isinstance(answer, (OutputParserException, Exception)):
+                    # If the agent encounters a parsing error or a server error after retries, replace the error
+                    # with a default value to prevent the pipeline from crashing
+                    results[i][j] = default_message
+                    logger.warning(
+                        f"Exception encountered in result[{i}][{j}]: {answer}. Replacing with default message.")
+
```
mdemoret-nv, line 91: This class can be used by any pipeline that wants to include agents. There should be nothing specific to the use case that were currently working on.

Instead, make this an option in the constructor of this class. Have a boolean, `replace_exceptions` which you can set to `True`/`False` to enable this functionality. And have a second value `replace_exceptions_value` that users can set to whatever they want to replace the exceptions with.

morpheus/llm/nodes/langchain_agent_node.py
```diff
@@ -78,6 +78,17 @@ async def execute(self, context: LLMContext) -> LLMContext:  # pylint: disable=i
...
+        # Processes the results to replace exceptions with a default message
+        default_message = "I do not have a definitive answer for this checklist item."
+        for i, answer_list in enumerate(results):
+            for j, answer in enumerate(answer_list):
+                if isinstance(answer, (OutputParserException, Exception)):
+                    # If the agent encounters a parsing error or a server error after retries, replace the error
+                    # with a default value to prevent the pipeline from crashing
+                    results[i][j] = default_message
+                    logger.warning(
+                        f"Exception encountered in result[{i}][{j}]: {answer}. Replacing with default message.")
```
mdemoret-nv, line 90: New functionality like this also needs a test to ensure its functioning.

mdemoret-nv (changes requested):
This needs to be generalized a bit more

morpheus/llm/nodes/langchain_agent_node.py
```diff
@@ -35,13 +37,22 @@ class LangChainAgentNode(LLMNodeBase):
...
 
         self._agent_executor = agent_executor
 
         self._input_names = self._agent_executor.input_keys
 
+        self._replace_exceptions = replace_exceptions
+        self._replace_exceptions_value = replace_exceptions_value
+
+        if self._replace_exceptions:
+            assert self._replace_exceptions_value is not None, "When replace_exceptions is enabled, replace_exceptions_value must be provided."
```
mdemoret-nv, line 54: `None` should be a perfectly acceptable value to replace exceptions with

mdemoret-nv (approved):

PR #1699, opened 2024-05-13
Add group by column stage

dagardner-nv:
## Description
* Adds new stage `GroupByColumnStage`


AnuradhaKaruppiah (commented):
Is there an issue# for this change? I would like to understand why it is being done.

dagardner-nv:
> Is there an issue# for this change? I would like to understand why it is being done.

The purpose of this stage is to ensure that the rest of the downstream stages in the pipeline get a unique message for each unique value in the column being grouped-by. 

The idea is that this would improve performance. In addition to this, if we know that each value in column 'A' has the same value, we can drop the column and store that value in the `ControlMessage.metadata` field.

I opened issue #1709 for this, but unfortunately I didn't create an issue prior to working on this feature.

morpheus/stages/preprocess/group_by_column_stage.py
```diff
@@ -0,0 +1,66 @@
...
+        return False
+
+    def on_data(self, message: MessageMeta) -> list[MessageMeta]:
+        with message.mutable_dataframe() as df:
+            grouper = df.groupby(self._column_name)
+
+        ouptut_messages = []
+        for group_name in grouper.groups.keys():
+            group_df = grouper.get_group(group_name)
+            ouptut_messages.append(MessageMeta(group_df))
```
AnuradhaKaruppiah, line 58: pls fix the spelling here.

morpheus/stages/preprocess/group_by_column_stage.py
```diff
@@ -0,0 +1,66 @@
...
+        self._column_name = column_name
+
+    @property
+    def name(self) -> str:
+        return "group-by-column"
+
+    def accepted_types(self) -> tuple:
+        return (MessageMeta, )
+
+    def supports_cpp_node(self) -> bool:
```
AnuradhaKaruppiah, line 48: @method?

morpheus/stages/preprocess/group_by_column_stage.py
```diff
@@ -0,0 +1,66 @@
...
+            grouper = df.groupby(self._column_name)
+
+        ouptut_messages = []
+        for group_name in grouper.groups.keys():
+            group_df = grouper.get_group(group_name)
+            ouptut_messages.append(MessageMeta(group_df))
+
+        return ouptut_messages
+
+    def _build_single(self, builder: mrc.Builder, input_node: mrc.SegmentObject) -> mrc.SegmentObject:
```
AnuradhaKaruppiah, line 62: Where is this private method being used?

AnuradhaKaruppiah:
LGTM, some minor suggestions.

morpheus/stages/preprocess/group_by_column_stage.py
dagardner-nv: Good catch, thanks!

morpheus/stages/preprocess/group_by_column_stage.py
dagardner-nv: The parent class `SinglePortStage` defines this, and will be invoked by the Pipeline (via `StageBase.build`) during the build phase.

AnuradhaKaruppiah (approved):

dagardner-nv:
test

AnuradhaKaruppiah (approved):

AnuradhaKaruppiah (approved):

PR #1716, opened 2024-05-21
Improve the logging tests and add support for resetting the logger

mdemoret-nv:
## Description
- Allows the logger to be "reset" which is necessary to avoid duplicate logging handlers if `configure_logging` is called multiple times
- Updates all of the logging tests to parameterize the log level and better check if the handlers are actually being called.


morpheus/utils/logger.py
```diff
@@ -112,49 +114,76 @@ def _configure_from_log_level(*extra_handlers: logging.Handler, log_level: int):
...
+
+
+def reset_logging():
+    """
+    Resets the Morpheus logging system. This will remove all handlers from the Morpheus logger and stop the queue
+    listener. This is useful for testing where the logging system needs to be reconfigured multiple times or
+    reconfigured with different settings.
+    """
+
+    morpheus_logger = logging.getLogger("morpheus")
```
dagardner-nv, line 172: I think the method should receive the logger name as a parameter, this would allow the unittests to use this to clear the root logger
```suggestion
def reset_logging(logger_name: str = "morpheus"):
    """
    Resets the Morpheus logging system. This will remove all handlers from the Morpheus logger and stop the queue
    listener. This is useful for testing where the logging system needs to be reconfigured multiple times or
    reconfigured with different settings.
    """

    morpheus_logger = logging.getLogger(logger_name)
```

morpheus/utils/logger.py
```diff
@@ -112,49 +114,76 @@ def _configure_from_log_level(*extra_handlers: logging.Handler, log_level: int):
...
+            handler.acquire()
+            handler.flush()
+            handler.close()
+        except (OSError, ValueError):
+            pass
+        finally:
+            handler.release()
+        morpheus_logger.removeHandler(handler)
+
+    delattr(morpheus_logger, "_configured_by_morpheus")
```
dagardner-nv, line 186: The unittests will need to call this method unconditionally, we should check to see if the attribute exists to avoid a potential `AttributeError`

```suggestion
    if hasattr(morpheus_logger, "_configured_by_morpheus"):
        delattr(morpheus_logger, "_configured_by_morpheus")
```

tests/test_logger.py
```diff
@@ -13,109 +13,168 @@
...
+    queue_listener_stop_mock.assert_called()
+
+
 @patch('logging.handlers.QueueHandler.emit')
-def test_configure_logging_from_level_default_handlers(queue_handler, queue_listener):
-    configure_logging(log_level=logging.DEBUG)
+@pytest.mark.parametrize("log_level", LogLevels)
+def test_configure_logging_from_level_default_handlers(queue_handler, log_level: type[LogLevels]):
+
+    reset_logging()
```
dagardner-nv, line 67: I would make this an autouse fixture.

dagardner-nv (approved):
LGTM, with a few changes the `clear_handlers` and `pytest_runtest_teardown` methods in `tests/conftest.py` could be updated to make use of the new `reset_logging` method.

PR #1809, opened 2024-07-11
Fix log parsing undefined variable and duplicate sequence id errors

### LINKED ISSUE #1765: [BUG]: undefined variable error in `examples/log_parsing/postprocessing.py`
### Version

24.06

Source

### Describe the bug.

In `LogParsingPostProcessingStage::__get_label_dicts` there is a possibility that the `new_label` and `new_confidence` variables are undefined. 

The code defines these inside an `if` block inside of a for-loop, and assumes one of two situations is true:
1) The `if` statement evaluated to `True` and are defined
2) The `if` statement evaluated to `False` but the variables were defined in a previous iteration of the loop.

There are situations where neither of the above are true and the code fails. 
This was encountered recently parsing a new dataset that was not in the format expected. 

We should either (or both):
* Define both variables outside the loop and handle these
* Detect that the input file does not match the format and raise a reasonable error

_No response_

### Relevant log output

<details><summary>Click here to see error details</summary><pre>

 [Paste the error here, it will be hidden by default]

</pre></details>

### Full env printout

<details><summary>Click here to see environment details</summary><pre>

 [Paste the results of print_env.sh here, it will be hidden by default]

</pre></details>

### Other/Misc.

_No response_

### Code of Conduct

dagardner-nv:
## Description
* Fixes two related bugs which occur when running the `log_parsing` example with unexpected data.
* Fix error in `LogParsingPostProcessingStage::__get_label_dicts` where there is a possibility that the `new_label` and `new_confidence` variables are undefined.

* Fix error where the `tokenize_text_series` method can return a sequence_ids list with duplicate ids when a text series is passed in and one of the values has a length longer than `seq_len` and `truncation=False`, this prevents a down-stream error in the inference stage
```
RuntimeError: Inconsistent ID column. Last element in 'seq_ids' tensor, [1], must not extend beyond last message, [0]
```
Not sure if this is the best place to catch this error.

Closes #1765


examples/log_parsing/postprocessing.py
```diff
@@ -122,17 +123,25 @@ def _postprocess(self, x: MultiResponseMessage):
...
+                if self._label_map[new_label] in token_dict:
+                    token_dict[self._label_map[new_label]] = (token_dict[self._label_map[new_label]] + " " + text_token)
+                else:
+                    token_dict[self._label_map[new_label]] = text_token
+
+                confidence_dict[self._label_map[label]].append(new_confidence)
             else:
-                token_dict[self._label_map[new_label]] = text_token
-            confidence_dict[self._label_map[label]].append(new_confidence)
+                warnings.warn(f"Ignoring unexecpected subword token: {text_token}")
```
mdemoret-nv, line 143: Using warnings will ensure this is only shown once, but does that mean once per function call, or once per message? We would want to ensure this gets called once per message so all the unexpected tokens were printed

tests/utils/test_cudf_subword_helper.py
```diff
@@ -0,0 +1,54 @@
...
+
+    with pytest.raises(ValueError):
+        tokenized = tokenize_text_series(vocab_hash_file=bert_cased_hash,
+                                         do_lower_case=do_lower_case,
+                                         text_ser=series,
+                                         seq_len=seq_length,
+                                         truncation=False,
+                                         stride=stride,
+                                         add_special_tokens=add_special_tokens)
+        print(tokenized.segment_ids)
```
mdemoret-nv, line 54: Remove print statements from tests. Use logging if necessary

mdemoret-nv (approved):

PR #1860, opened 2024-08-23
Revert PR_1736  "Always run the PR builder step even if others are cancelled"

AnuradhaKaruppiah:
CIs are failing because of #1736 not playing well with  https://github.com/rapidsai/shared-workflows/pull/237

As a stopgap I am reverting the changes in #1736


dagardner-nv (approved):
Thanks for fixing this

PR #1878, opened 2024-08-29
Implement `MultiProcessingStage`

### LINKED ISSUE #1850: [FEA]: Add Multi-Processing Stage

New Feature

Critical (currently preventing usage)

Many users have created versions of stages which create a multi-process pool and farm work out pool as they arrive. The added benefit here is the ability to scale partial pieces of the pipeline to work around the GIL and increase performance.

Create a new stage which will unify the many different implementations into a well tested and flexible design. It should support the following:
- Processing messages with multiple processes
- Correctly handling backpressure
- Ability to run multiple instances of the stage in the same pipeline without collisions
- Easy to setup
- Work without any external packages like dask or ray

### Additional context

_No response_

### Code of Conduct

yczhang-nv:
## Description

Implemented `SharedProcessPool` and `MultiProcessingStage`.

**SharedProcessPool**
  - A singleton process pool that manages tasks from multiple stages, each with its own queue. Allows setting stage-specific process usage to cap the number of workers a stage can utilize.
  - The total number of workers is determined by the environment variable `SHARED_PROCESS_POOL_CPU_USAGE`, which is multiplied by the system’s available CPU cores (`os.sched_getaffinity(0)`). 
    - If the variable is not set, the default is 50% of available CPUs.
  - Accepts process function with arbitrary argument signatures.

**MultiProcessingStage**
  - A versatile base stage that uses SharedProcessPool to handle incoming messages in a multiprocessing manner.
  - Accepts processing functions with a single argument, capable of handling dynamic input and output types. The types are deducted at runtime based on the provided processing function.
  - Two ways to use the stage:
    - Extend `MultiProcessingBaseStage` and implement the data processing logic in the `_on_data()` method.
    - Use `MultiProcessingStage.create()` by passing in the processing function directly.
  - Able to run multiple instances of the stage in the same pipeline. 

Closes #1850 


tests/utils/test_shared_process_pool.py
```diff
@@ -0,0 +1,291 @@
...
+        tasks.append(pool.submit_task("test_stage_2", _time_consuming_task, sleep_time))
+        tasks.append(pool.submit_task("test_stage_3", _time_consuming_task, sleep_time))
+
+    pool.stop()
+    pool.join()
+
+    # all tasks should be completed before the pool is shutdown
+    assert len(tasks) == 3 * task_num
+    for task in tasks:
+        assert task.done()
```
mdemoret-nv, line 270: Instead of using sleep which is terrible for testing, you can exercise the same functionality like this:

```suggestion
def block_until_signaled(q: multiprocessing.Queue):
    return q.get()


def test_task_completion_with_early_stop(shared_process_pool: SharedProcessPool):

    pool = shared_process_pool
    pool.set_usage("test_stage_1", 0.1)
    pool.set_usage("test_stage_2", 0.3)
    pool.set_usage("test_stage_3", 0.5)

    m = multiprocessing.Manager()
    queue = m.Queue()

    tasks = []

    task_num = 5
    sleep_time = 10
    for _ in range(task_num):
        tasks.append(pool.submit_task("test_stage_1", block_until_signaled, queue))
        tasks.append(pool.submit_task("test_stage_2", block_until_signaled, queue))
        tasks.append(pool.submit_task("test_stage_3", block_until_signaled, queue))

    pool.stop()

    # Ensure that all tasks have not completed since they havent been signaled yet
    for task in tasks:
        assert not task.done()

    # Signal the tasks to complete
    for i in range(len(tasks)):
        queue.put(i)

    pool.join()

    # all tasks should be completed before the pool is shutdown
    assert len(tasks) == 3 * task_num
    for task in tasks:
        assert task.done()
```

Note: This does require some changes to how `stop` and `join` function. See comments on those functions

python/morpheus/morpheus/utils/shared_process_pool.py
```diff
@@ -0,0 +1,409 @@
...
+        # no new tasks will be accepted from this point
+        self._status = PoolStatus.STOPPING
+
+        # wait for all task queues to be empty
+        task_queue_count = len(self._task_queues)
+        empty_task_queues = set()
+        while len(empty_task_queues) < task_queue_count:
+            for stage_name, task_queue in self._task_queues.items():
+                if task_queue.empty():
+                    empty_task_queues.add(stage_name)
```
mdemoret-nv, line 334: WHat is this block for? I cant see where `empty_task_queues` is used

python/morpheus/morpheus/utils/shared_process_pool.py
```diff
@@ -0,0 +1,409 @@
...
+
+        self._cancellation_token.cancel()
+
+        for i, p in enumerate(self._processes):
+            p.join()
+            logger.debug("Process %s/%s has been joined.", i + 1, self._total_max_workers)
+
+        logger.debug("SharedProcessPool.stop(): All tasks have been completed. SharedProcessPool has been stopped.")
+        self._status = PoolStatus.SHUTDOWN
```
mdemoret-nv, line 343: Stop should be non-blocking. `stop()` means, do not accept any more work, let everything finish naturally. `join()` is the blocking function which means, wait unti everything is cleaned up and complete.

python/morpheus/morpheus/utils/shared_process_pool.py
```diff
@@ -0,0 +1,409 @@
...
+        ------
+        TimeoutError
+            If has been waiting more than the timeout.
+        """
+        start_time = time.time()
+
+        while self._status != PoolStatus.SHUTDOWN:
+            if timeout is not None and time.time() - start_time > timeout:
+                raise TimeoutError("SharedProcessPool join has timed out.")
+            time.sleep(0.1)
```
mdemoret-nv, line 404: We need to improve this join method. A better way would be to use some other synchronization primitive. Usually, a condition variable with a predicate works well here. Let me know if you need examples. When using a condition variable, you will need to spawn a new thread which does something like:
```python
for p in self._processes:
   p.join()

self._status = PoolStatus.SHUTDOWN

self._condition_variable.notify_all()
```

mdemoret-nv (changes requested):
Very close. Need a few tweaks to how stop/join works. Also, can you add more type annotations to this PR? For any public function, type annotations are helpful.

tests/utils/test_shared_process_pool.py
```diff
@@ -277,15 +295,20 @@ def test_terminate_running_tasks(shared_process_pool):
...
+        tasks.append(pool.submit_task("test_stage_2", _blocked_until_signaled_task, queue))
+        tasks.append(pool.submit_task("test_stage_3", _blocked_until_signaled_task, queue))
 
-    # The pool should be shutdown immediately after calling terminate() without waiting for the tasks to complete
     pool.terminate()
-    pool.join()
+
+    # No tasks have been completed since they have not been signaled yet
+    for task in tasks:
+        assert not task.done()
```
mdemoret-nv, line 314: After `terminate` is called, we want to still call `pool.join()`. Its ok if some tasks have finished here. We just want to assert that not ALL tasks finished.

python/morpheus/morpheus/utils/shared_process_pool.py
```diff
@@ -115,20 +113,25 @@ def __new__(cls):
...
-                logger.info("SharedProcessPool.__new__: instance already exists.")
+                logger.debug("SharedProcessPool.__new__: instance already exists.")
 
         return cls._instance
 
     def _initialize(self):
         self._status = PoolStatus.INITIALIZING
 
-        self._total_max_workers = math.floor(max(1, len(os.sched_getaffinity(0)) * 0.1))
+        cpu_usage = os.environ.get("SHARED_PROCESS_POOL_CPU_USAGE", None)
```
mdemoret-nv, line 129: Prefix all environment variables with `MORPHEUS_`

mdemoret-nv (approved):
Much better implementation with the condition variables. Looks good.

PR #1883, opened 2024-09-05
Scripts for building and uploading the morpheus-core conda package

AnuradhaKaruppiah:
With this change two conda packages are generated -
1. morpheus - with all the components (core + llm + examples + tests). This is existing code.
2. morpheus-core - with only the core (sources, stages and sinks)
The morpheus-core package is uploaded to anaconda.org/nvidia for use by applications.

A new PR label "conda-core-build" is available to trigger building the morpheus-core conda package as a part of the CI pipeline.

Testing done:
Generated the morpheus-core conda package via  -
a. local ci runner -
```
ci/scripts/run_ci_local.sh conda_core
```
b. docker build -
```
DOCKER_IMAGE_TAG=latest DOCKER_IMAGE_NAME=conda_bld_morpheus_core DOCKER_TARGET=conda_bld_morpheus_core DOCKER_EXTRA_ARGS="--build-arg CONDA_UPLOAD=upload --build-arg CONDA_TOKEN=<my-token> --progress plain" ./docker/build_container.sh
```
And verified an application that uses the core stages.


.github/workflows/ci_pipe.yml
```diff
@@ -21,8 +21,16 @@ on:
       run_check:
         required: true
         type: boolean
+      conda_core_run_build:
+        description: 'Runs the stage to build the morpheus-core conda package'
+        required: true
+        type: boolean
+      conda_core_upload_label:
```
dagardner-nv, line 28: I think the `conda_upload_label` should apply to all conda packages being built. 

I don't imagine there would be an instance where we build conda packages for morpheus_core and morpheus_llm, and upload one but not the other.

.github/workflows/ci_pipe.yml
```diff
@@ -21,8 +21,16 @@ on:
       run_check:
         required: true
         type: boolean
+      conda_core_run_build:
```
dagardner-nv, line 24: I'm thinking in most cases if we're building conda packages in CI we would want to build all of them.

.github/workflows/pr.yaml
```diff
@@ -89,13 +90,18 @@ jobs:
...
+      conda_core_run_build: ${{ !fromJSON(needs.prepare.outputs.is_pr) || fromJSON(needs.prepare.outputs.has_conda_core_build_label) }}
+      # Upload morpheus-core conda package only for non PR branches. Use 'main' for main branch and 'dev' for all other branches
+      conda_core_upload_label: ${{ !fromJSON(needs.prepare.outputs.is_pr) && (fromJSON(needs.prepare.outputs.is_main_branch) && 'main' || 'dev') || '' }}
+      # Run morpheus conda package, with all components. This is done for main/dev
+      # branches and for PRs with the conda-build label.
       conda_run_build: ${{ !fromJSON(needs.prepare.outputs.is_pr) || fromJSON(needs.prepare.outputs.has_conda_build_label) }}
-      # Update conda package only for non PR branches. Use 'main' for main branch and 'dev' for all other branches
       # Disable conda upload for now, once we have morpheus packages in conda forge set the value to
-      # !fromJSON(needs.prepare.outputs.is_pr) && (fromJSON(needs.prepare.outputs.is_main_branch) && 'main' || 'dev') || ''
       conda_upload_label: ""
```
dagardner-nv, line 102: Shouldn't this be set to the same as `conda_core_upload_label`?

dagardner-nv (approved):
LGTM, just a few comments/questions

.github/workflows/ci_pipe.yml
AnuradhaKaruppiah: Agreed. I will make that tag more generic.

.github/workflows/ci_pipe.yml
AnuradhaKaruppiah: I agree with that and I was tempted to use the same label but it also seemed useful to create one for "core" as that is likely going to be the code that is updated most often.

I will revisit this when the next conda package is created. There may not be much value in proliferating labels.

.github/workflows/pr.yaml
AnuradhaKaruppiah: Do we need to start uploading the morpheus package that includes all components? I created the separate upload label for the morpheus-core package mainly to hold this one back.

AnuradhaKaruppiah: Based on further discussion, we will not upload the "morpheus" conda package because of the pip dependencies.

.github/workflows/ci_pipe.yml
AnuradhaKaruppiah: I have dropped the new upload label and consolidated it with `conda_upload_label`.
In the process need to update the job for the "morpheus" conda build to prevent it from accidentally uploading that package.

PR #2090, opened 2024-12-19
Avoid private cudf DeviceScalar in favor of using pylibcudf & pyarrow

mroeschke:
## Description
In cudf, I'm looking to ultimately remove the `DeviceScalar` implementation in favor of using `pylibcudf.Scalar` and `pyarrow`.

Before, it appears there was a cudf helper from getting an element from a string column's child (offset) column. There now exists pylibcudf APIs to do this and therefore can avoid having to directly access DeviceScalar

### DIFF
--- a/python/morpheus/morpheus/_lib/cudf_helpers/__init__.pyi
+++ b/python/morpheus/morpheus/_lib/cudf_helpers/__init__.pyi
@@ -8,6 +8,7 @@ from cudf.core.dtypes import StructDtype
 import _cython_3_0_11
 import cudf
 import itertools
+import pylibcudf
 import rmm
 
 __all__ = [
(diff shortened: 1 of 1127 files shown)


python/morpheus/morpheus/_lib/cudf_helpers/__init__.pyi
```diff
@@ -19,6 +20,7 @@ __all__ = [
     "bitmask_allocation_size_bytes",
     "cudf",
     "itertools",
+    "plc",
```
dagardner-nv, line 23: I think this is a type-o
```suggestion
    "pylibcudf",
```

dagardner-nv (approved):

PR #2109, opened 2025-01-15
Remove some cudf._lib.column.Column annotations in Cython

mroeschke:
## Description
Related to https://github.com/rapidsai/cudf/issues/17317, `cudf._lib.column.Column` is being phased out in favor of using pylibcudf.

I found some locations where `cudf._lib.column.Column` was used as type annotations in Cython; however, the accessed or returned objects were calling Python defined attributes/methods or used as objects in other Python objects respectively. Therefore, I don't suspect the annotations are entirely performance critical here

### DIFF
--- a/python/morpheus/morpheus/_lib/cudf_helpers.pyx
+++ b/python/morpheus/morpheus/_lib/cudf_helpers.pyx
@@ -28,8 +28,6 @@ from pylibcudf.libcudf.io.types cimport table_with_metadata
 from pylibcudf.libcudf.table.table_view cimport table_view
 from pylibcudf.libcudf.types cimport size_type
 
-from cudf._lib.column cimport Column
-
 ##### THE FOLLOWING CODE IS COPIED FROM CUDF AND SHOULD BE REMOVED WHEN UPDATING TO cudf>=24.12 #####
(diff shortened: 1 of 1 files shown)


python/morpheus/morpheus/_lib/cudf_helpers.pyx
```diff
@@ -61,7 +59,7 @@ from cudf._lib.types cimport (
     # dtype_to_pylibcudf_type,
 )
 from cudf._lib.null_mask import bitmask_allocation_size_bytes
-
+from cudf._lib.column cimport Column
```
dagardner-nv, line 62: Is this import still needed, and out of curiosity is there a reason why it was moved?

mroeschke: Yup, it appears this is still used in `from_column_view_with_fix`, so I moved this between the `##### THE FOLLOWING CODE IS COPIED FROM CUDF AND SHOULD BE REMOVED WHEN UPDATING TO cudf>=24.12 #####` comment so it can be removed when the upgrade occurs

dagardner-nv (approved):

PR #2215, opened 2025-05-08
Fix script reference in DFP example's Dockerfile from 125 to 128

Misiac:
## Description
Fixes the DFP example not building due to the Dockerfile referencing an outdated Conda environment file.

Updated the environment file reference from `dfp_example_cuda-125_arch-$(arch).yaml` to `dfp_example_cuda-128_arch-$(arch).yaml` to match the currently available config.


dagardner-nv (approved):
Good catch, thanks for the PR.

PR #2257, opened 2025-07-14
Fix handling of labels in DLP example

dagardner-nv:
## Description
* The labels have to match those in the `labels.pt` file exactly, currently they are a subset.
* In addition to this, the onnx model is unable to generate the labels embedding.
* Disable input sampling by default

### DIFF
--- a/examples/data_loss_prevention/dlp_stages/gliner_triton.py
+++ b/examples/data_loss_prevention/dlp_stages/gliner_triton.py
@@ -39,6 +39,7 @@ class GliNERTritonInference:
         Path to the directory containing the GLiNER model files. Used for pre and post-processing.
     labels : list[str]
         List of entity labels to detect, this should match the named patterns used in the RegexProcessor stage.
+        Ignored if the `label_embedding.pt` file is present in the `model_source_dir`.
     onnx_path : str, default = "model.onnx"
         Path to the ONNX model file, relative to the `model_source_dir`. Default is "model.onnx".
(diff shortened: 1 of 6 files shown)


tzemicheal (approved):
The changes in the run config, and fix to triton GLiNER loading LGTM!

examples/data_loss_prevention/dlp_stages/gliner_triton.py
```diff
@@ -112,9 +114,9 @@ def _load_label_data(self):
             label_data = torch.load(self._labels_file)
 
             labels_embeddings = label_data['embeddings']
-            if sorted(self._labels) != sorted(label_data['labels']):
-                raise ValueError("Label mismatch between model and label file.")
+            self._labels = label_data['labels']
         else:
+            self._labels = self._user_labels
```
willkill07, line 119: (Just line 119) Do we want to have any sort of validation here? Just permitting an arbitrary list[str] may not be desirable. Or does calling `model.encode_labels` perform the validation?

willkill07 (approved):

examples/data_loss_prevention/dlp_stages/gliner_triton.py
dagardner-nv: There isn't much validation that this stage is able to perform. The only real requirement for the labels is that they match the keys of the regex patterns that is the type of pii we are searching for ('ssn', 'credit_card'...)

PR #2277, opened 2025-07-17
Fix: examples build on aarch64

willkill07:
There was a linker error on aarch64 indicating that `-lmorpheus` could not be found.

Two possible solutions existed:
1. change the LINK_TARGETS from `morpheus` to `morpheus::morpheus`
2. add `find_library` calls to find the morpheus shared library on the system.

(1) unfortunately resulted in a public link failing of mrc::pymrc -- which probably indicates that we shouldn't be exposing that as a public link dependency.

(2) was chosen because it's what we'd expect the system to do -- find the installed morpheus shared object file and link to it.

It's a bit cumbersome but it works as intended. This is more of a hotfix/patch rather than a complete engineered solution. Improvements and enhancements are welcome.

## Description

Closes 


dagardner-nv (approved):
Thanks!

PR #2294, opened 2025-07-23
Document the source of the gliner model

dagardner-nv:

### DIFF
--- a/models/dlp_models/gliner_bi_encoder/attribution.md
+++ b/models/dlp_models/gliner_bi_encoder/attribution.md
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:4dbfe4b8285f816e8595ec6e569a9c83a38d035ae25d132864316a94a4fab4f9
+size 1222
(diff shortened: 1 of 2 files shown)


models/.gitattributes
```diff
@@ -7,7 +7,7 @@
 abp-models/* filter=lfs diff=lfs merge=lfs -text
 datasets/training-data/** filter=lfs diff=lfs merge=lfs -text
 datasets/validation-data/** filter=lfs diff=lfs merge=lfs -text
-dlp_models/gliner_bi_encoder/** filter=lfs diff=lfs merge=lfs -text
+dlp_models/gliner_bi_encoder/*.json filter=lfs diff=lfs merge=lfs -text
```
hsin-c, line 10: Not a blocker, but I noticed that the naming convention of the models/ directory is using `-` (so this folder would be `dlp-models` instead.

hsin-c, line 10: Have we decided not to track the ONNX model with LFS?

hsin-c (approved):

willkill07 (approved):

